Issue #1164, opened 2015-02-19
Debugging inside OS callbacks


wuryel:
I would like to extend the mbeddr debugger to my extension of mbeddr. The "new" DSL shall provide access the the tasking facilities of the OS below. Right now I focus on a Linux host and use the pthread_create(...) directive to create and start a concurrent thread. This primitive expects a pointer
to a code snippet which is executed as a separate thread. Essentially the OS calls this code snippet whenever the thread is being scheduled.

Now I can place a breakpoint on DSL level into this code snippet and the debugger holds correctly but the stack frame construction fails when lifting the debugger information to DSL level.

The call stack information provided by the debugger goes below the DSL/Application level (so internals of the pthread_create(...) primitive are reported). This information (obviously) can't be mapped to the respective node on MPS tree level and the whole thing crashes.

I did some investigation in the mbeddr debugger implementation and identified a few checks that I removed in order to be able to successfully debug.

In the current implementation exceptions are thrown if it is not possible to map debugger information to MPS nodes. After my tiny changes these exceptions are not thrown anymore, but there is just no mapping performed. Maybe this is not the intended or best way to solve the issue (by far I haven't understood all the implementation), but at least it allows debugging code inside an OS callback routine.

This is what I did:

1) Source files refered to by the debug info are allowed to have the extension ".S" and ".s" besides just ".c". So this had to be added in these three operations

1a)  -- com.mbeddr.core.debug.(util models).debugger.(mapper).(state).(frame-mapper).(impl).MFrameMapperImpl.getFileName()
1b)  -- com.mbeddr.core.debug.(util models).debugger.(mapper).(state).(frame-mapper).(impl).GeneratedCFilePath.GeneratedCFilePath()
1c)  -- com.mbeddr.core.debug.(util models).debugger.(mapper).(state).(frame-mapper).(impl).GeneratedCFilePath.getUnitName

2) If the stack frame tracing doesn't work, this doesn't necessarily mean that we have an issue. Maybe a warning is sufficient here, so the user is aware of the fact.
To solve this I just removed the statements throwing the respective exceptions:

2a)  -- com.mbeddr.core.debug.(util models).debugger.(mapper).(state).(frame-mapper).(impl).MFrameMapperImpl.getClosestNodeViaTracesNew()
    - within the else block the statement "throw new MMissingTracesException();" is removed;
    - instead "return null;" is inserted

[...]

PR #1165, opened 2015-02-25
Debugging inside OS callbacks #1164

wuryel:
ok, I hope I have done this correctly now ;-)

please also see the second commit for (probably) a copy'n'paste issue


DomenikP:
merged.

PR #1177, opened 2015-05-08
Suggestions

vinnim:
null (NULL) can be explicitly (void*)0 instead of 0. Invisible with gcc, but makes cbmc on windows aware of null pointers and all (with flag --pointer-check).

Misra C 2004 Sec. 6.11 refers to NULL only as the "null pointer constant (the value 0 cast to type void*)". 
However, C11 Sec. 6.3.2.3 paragraph 3 says that "an integer constant expression with the value 0, or such an expression cast to type void *, is called a null pointer constant" -- thus also covering the 0 constant without cast.


coolya:
@stephaneberle9 can you test if this works with your embedded compilers?

@danielratiu are you fine with this?

danielratiu:
this change makes cbmc analyses work better and is conform to MISRA

coolya:
Looks good. I'll merge.

PR #1187, opened 2015-06-11
Concurrency analyses

vinnim:
Suggestion


danielratiu:
as discussed with Vincent

PR #1218, opened 2015-08-07
updated the plantuml jar

farcasia:
As suggested on the mbeddr forum, I have updated the plantuml jar file to the newest version available on the plantuml website and made a pull request.


coolya:
I'm looking into it right now. Looks like we have lots of duplication inside of our repo. I will fix that while merging your request for just having a single plantuml jar in the repo.

PR #1296, opened 2016-02-04
extend documentation plugin solution to add concept documentation in …

palpandi111:
Hi,

We did some changes to extend the documentation plugin solution to add concept documentation in documentation language. Documentation language is a language with only documentation aspect.
Previously we should add documentation for the concept in the same language.

To enable this we have added few changes in documentation plugin solution.
There we created a singleton class to provide documentation languages by filtering languages with only documentation aspect from all languages.
Then we are showing the documentation for the concept from the documentation language if available.

Regards,
Palpandi

### DIFF
--- a/code/languages/com.mbeddr.doc.aspect/languages/com.mbeddr.doc.aspect/models/com/mbeddr/doc/aspect/constraints.mps
+++ b/code/languages/com.mbeddr.doc.aspect/languages/com.mbeddr.doc.aspect/models/com/mbeddr/doc/aspect/constraints.mps
@@ -124,6 +124,9 @@
       <concept id="1171315804604" name="jetbrains.mps.lang.smodel.structure.Model_RootsOperation" flags="nn" index="2RRcyG">
         <reference id="1171315804605" name="concept" index="2RRcyH" />
       </concept>
+      <concept id="1176109685393" name="jetbrains.mps.lang.smodel.structure.Model_RootsIncludingImportedOperation" flags="nn" index="3lApI0">
+        <reference id="1176109685394" name="concept" index="3lApI3" />
+      </concept>
       <concept id="1139613262185" name="jetbrains.mps.lang.smodel.structure.Node_GetParentOperation" flags="nn" index="1mfA1w" />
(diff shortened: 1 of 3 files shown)


coolya:
There is a technical and a conceptual problem with this implementation:

- The DocHelper is using a static list to cache it's result. That will will break if there more than one open project.
- Looking into all languages which are in the current project will become a performance problem once more mbeddr languages start using the doc aspect.

I'm also not in favor of packing this behavior into the UI. There might be other places where we want to use that documentation. I think delegation of documentation should be hidden behind the documentation descriptor. I would prefer putting a special root in the documentation aspect of a language that then delegates the documentation to some other place. That way searching for documentation will not be slower if more languages are loaded in the project.

Putting @wsafonov in the loop because he has written the documentation aspect.

wsafonov:
I would agree with Kolja especially regarding the possible performance issue with looking into all languages and their documentation aspects. A better approach would be to reduce the set of relevant languages to only those that extend from the language of the node we currently inspect for documentation. That is if somebody wants to document concepts from some other language, he would need to extend it from the created language.

balakrishnan16:
Hi,
I agree with both of your comments. We have already moved out of development phase, just deviated the process for this task alone. Can we have this implementation for now and we will add a task on ourselves to have a improvised version of the same in next immediate release? 

Regards
Balakrishnan

wsafonov:
I have actually already written a enhancement for the concept documentation lookup to consider all extending languages. Now when writing a documentation in a new language, you will be able to assign a documentation section to any concept from your language or any language extended by it.

DineshLMS:
Hi All,

Thank you very much for explaining us the possible performance issues and thanks for providing us some enhancements also. I do agree with your points.

I went through the code of mbeddr.doc.aspect.ui.plugin and wrote some enhancements such that even if the documentation aspect of the current language is null, it will look for the documentation aspects in the language that extends the current language. 

With this way, we need not scan through all the languages in the registry and will not result in performance issues like our previous implementation.

I will send a seperate pull request ASAP. Please review the same and merge the same.

Regards,

Dinesh

PR #1316, opened 2016-03-01
Some minor fixes

wuryel:
I have come across some minor issues lately. Please have a look.


coolya:
Huge thanks for all of these additions and fixes. :+1:  
I comment on all the individual issue you have created.

Since there are many things I would merge without any further changes to them it might be worth to split this pull request. That way we can merge them fast and work on the remaining things together. What do you think?

wuryel:
Ok, how do I split a pull request?
Can't you pick certain commits?

wuryel:
oops - wrong button

coolya:
Yes, I can also just pick the commits one by one.

PR #1323, opened 2016-03-02
Feature/assertequals

swathikrishnamoorthy:
Added overridden code location To assert Equals for structure type to get link, if assert fails


coolya:
Can you please sent that pull request against the milestone branch not against master?

PR #1325, opened 2016-03-03
Adding overridden code location and context node & model id to assert…

swathikrishnamoorthy:
Added overridden code location to assert Equals For struct Type.


DomenikP:
merged :+1:

PR #1423, opened 2016-07-11
concurrency language fixes

DineshLMS:
Hi,

	tasks refered from instances lying several levels deeper inside the composite component is referred properly now.

Please review and accept the request.
Please ensure no tests are broken.
I didnt run unit tests,

Regards, 
Dinesh


coolya:
Can you please provide a unit tests with a sample that would fail without this modification?

I would like to have a test to ensure that we don't break this use case in the future.

DineshLMS:
Hi Kolja,

I have added an executable test.

Please review the same.

Regards,

Dinesh

coolya:
Looks good to me

PR #1481, opened 2016-09-08
allow overridding config items

DineshLMS:
Hi,

Is it possible to allow the overriding of config items. In case, If I dont want to show much information to the user, I will override the config item and provide a new editor for the same but the all the details that needs to be inputted will be programatically added in the overridden config item

Regards,
Dinesh


coolya:
@DomenikP haven't we implemented this?

DomenikP:
yes, the change was merged yesterday from the branch refactor/InheritanceInConfigItems into milestone/v15-2

PR #1574, opened 2016-11-09
Illegal model access error

DineshLMS:
Hi,

Several exceptions are thrown in the console when the ESD/mbeddr project is opened. One such is the illegal model access error saying that you can write models only in write action from c.m.core.base.pluginsolution. I am attaching the stacktrace as a text file here. 
[stacktrace.txt](https://github.com/mbeddr/mbeddr.core/files/580039/stacktrace.txt)

The stack trace navigates to the  "PreferenceModules" class file where createModule is done inside a read action. I have changed the read action to write action and also added a write action where similar operation is done.

Please review and merge into milestone/v15-3

Regards,

Dinesh


coolya:
Haven't we fixed this on master some time ago? @DomenikP

DomenikP:
I cherry picked the fix from the master branch onto the milestone/v15-3.

PR #1578, opened 2016-11-11
add mbeddr publication to build against complete mbeddr

coolya:
I had to add a new publication that contains all the individual zips of the mbeddr plugins in order to able to build other projects against mbeddr that depend of more than the platform.

@sergej-koscejev can you have look that I didn't screw up anything else with it?


sergej-koscejev:
Postponed until Monday after private discussion with @coolya.

PR #1584, opened 2016-11-22
Framework to contribute the reserved keywords missing in IIdentifierNamedConcept

DineshLMS:
Hi,

If the ESD or other customers who inherits the mbeddr.core code wants to contribute a set of keywords that the user should not be able to use as identifier, the framework for the same is not available in mbeddr.core. The issue is DineshLMS/mbeddr.core#10 [just for the reference]

It is a simple fix which introduces a behaviour method and made the IIdentifierNamedConcept to check for the behaviour method contents which can be overridden by the its child concepts.

Regards,
Dinesh


DineshLMS:
This approach will work only for specific concepts and not for all instances of IIDentifierNamedConcept. I will come up with some other implementation

PR #1610, opened 2017-01-12
Creating connections between component instances inside composite components

Nandha1712:
Hi,
         Modified the code so that the proper connections between components can be given in graphicalWiring projection of Composite Components. Earlier we were able to create the connections but they were incomplete connections. 
example for pull request

So please review the code and merge it with milestone/v16-1


slisson:
There were several problems in the components diagram. I fixed them on master (84b058c53d6319a9f96896f99e7b7c54686478c0) and milestone/v16-1 (cc62381f201dbad8228768c306e1174c20818a18).

Issue #1611, opened 2017-01-12
Components not appearing in diagram palette when Composite Components is selected in graphicalWiring projection


Nandha1712:
1. I have 2 components in the implementation module.
2. I created a Composite Component.
3. In graphicalWiring projection, the Composite Component changes to diagram view.
4. The name of the Components should come in the diagram palette.
4. But they are not coming in the palette, though code is written to provide the contents for diagram palette in the **_graphicalWiring Editor of CompositeComponentInstanceConfig._**

Current scenario:
![for reporting bug](https://cloud.githubusercontent.com/assets/10434575/21882760/f9c4a37e-d8d1-11e6-9512-c4b24f497031.jpg)

Expected Result:

![correct palette](https://cloud.githubusercontent.com/assets/10434575/21882777/0b9bb9fc-d8d2-11e6-9a1b-e1c7b61b5bf4.jpg)

PR #1612, opened 2017-01-12
fix/workaround for issue-In graphicalWiring projection, Components not appearing in diagram palette when composite component is selected

Nandha1712:
1. Created the necessary _SubstituteInfo_ manually in palette entries section since the Components are not available in diagram palette.
2. It is a fix/work around for the issue [1611](https://github.com/mbeddr/mbeddr.core/issues/1611)
3. Please review the code and merge with milestone/v16-1 if this fix seems to be a valid one.


slisson:
There were several problems in the components diagram. I fixed them on master (84b058c53d6319a9f96896f99e7b7c54686478c0) and milestone/v16-1 (cc62381f201dbad8228768c306e1174c20818a18).

PR #1615, opened 2017-01-18
platform: fix NPE caused by InterpreterBase

coolya:
InterpreterBase was not passing a instance of NullCoverageAnalyzer but null
this cause the evaluators to throw NPEs.


coolya:
As there is no public API change and minor fix, merged without additional review

PR #1616, opened 2017-01-20
Feature/enum can mangle names

DineshLMS:
Hi,

This pull request is related to #1609. 

I have implemented the ICanMangleNames interface for enum literals and changed the mangled name behaviour method accordingly.

Regards,

Dinesh


DineshLMS:
A config item is always an additional work for the user. To solve the problem that if the enum is present in external module, we should not mangle names and if the enum is present in impl.module, we can choose to mangle names (if duplicate names present). I will update this pull request by making the textgen aspect intelligent enough to handle these cases.

DineshLMS:
Modified the *mangledname()* behaviour method of enum literal... If the enum literals are in external module, the method returns the name of the literal. If prevent name mangling is on, the method returns the name of the literal. Otherwise, mangledNameOfEnumDecl___enumLiteralName is returned.

coolya:
One note here, please override the `preventNameMangling` method in the EnumLiteral instead of putting the logic in the name method.

DineshLMS:
can you please check this and tell me what more needs to be done

coolya:
The pull request still contains a lot of commits, I think there is something wrong with it. 

Could you please rebase you commits in the milestone branch so that it can be merged with a single commit?

DineshLMS:
sure

PR #1644, opened 2017-03-07
Milestone/v16 1

aswinsugumaran:
1. Value is rounded off when its is a double used in the assert statements.(UnitTestHelper) No method for the range value when its a double type.
Exp value = 36.99999
value shown = 36
2. Need provision to specify for values in ranges. Say a < 0 <b.


DomenikP:
wrong target branch.

Issue #1646, opened 2017-03-08
Cannot use a predefined variable in For loop initialization


aswinsugumaran:
Hello, 
       I have a variable declared outside the for loop.When i try to reuse the same variable again in initialization part of the for loop its not visible. I have tried this with some other loops statements which works fine. Only for the For loop it doesn't seem to work only at the initialization part.Any suggestion would be of great help.

![forloop](https://cloud.githubusercontent.com/assets/25502999/23700108/da50f260-0417-11e7-90c2-230827b54dd3.png)

aswinsugumaran:
a

aswinsugumaran:
@coolya  from what i inferred we have this constraint named ForVarRef_Constraints which returns false cause of iterator is null at time of for loop initialization. I set the constraint to be true manually which works now. Is this the proper solution for this or any other alternate solution is there?

coolya:
I will have a look at this tomorrow. From you error description I don't really understand what the issue is. I guess `ix` is not in scope, right?

Can give a text example what you want code you would like to write?

aswinsugumaran:
@coolya  the ix is in the scope. 
Steps to reproduce 
1. Create a local variable
2. Try to use the local variable in the simple for loop
3. The local variable will not be available in the code completion menu of for loop initialization part.

![forloop1](https://cloud.githubusercontent.com/assets/25502999/23750878/eeb51226-04f4-11e7-84ba-fcc6e35a26d4.png)

coolya:
What you do at that location is defining the iterator. There is no way to use `ix` at that location. 

Once again, please give me a **text**  example of the code you would like to be able to enter.

aswinsugumaran:
Example
int a;

for( <the variable 'a' should be listed here when i use the code completion menu > ;    ; )

coolya:
I don't think this makes much sense. According to the C99 spec 6.8.5.3 a for statements is defined like this:

```for ( clause-1 ; expression-2 ; expression-3 ) statement```

Where clause-1 can either be a declaration, which would make it a variable with a scope within the for loop, or a void expression.

The void expression case is not supported by mbeddr, you are only allowed to declare a variable at that location. 

What should work is something like 
```

int16 a;

for (int16 i = a; a++ ;) {
}

```

DineshLMS:
[...]

PR #1661, opened 2017-03-24
For statement: provision to use predefined local variable in for loop initializer

aswinsugumaran:
Hi,
https://github.com/mbeddr/mbeddr.core/issues/1646
 
As stated in the issue and we cannot use a locally defined variable in for loop initializer, so i have removed the ForVarRef constraint to enable those variable in the for loop initializer.

Please review and merge into milestone/v16-1

Regards,
Aswin Sugumaran


DomenikP:
I don't see why we want to enable the initialisation of local variables in for statement headers, since this is not supported by the C99 standard. Introducing this possibility will break mbeddr's C99 compatibility (see Kolja's comment in the linked issue: https://github.com/mbeddr/mbeddr.core/issues/1646).

To enable this feature I could think of introducing a transformation in the statements generator that pulls out such initialisations from a for-statement and creates new statements before this for statement that perform these initialisations.

DomenikP:
I discussed with Aswin that we will keep the language implementation as it is, but provide a migration script that 1.) pulls out instances of ForVarRef (used to init. local variables in ForStatement headers) and 2.) creates assignment expressions before the ForStatement that perform the initializations.

PR #1701, opened 2017-05-19
Removal of Deprecated API string_tokenizer

DineshLMS:
Hi,

Analysis code was trying to get additional paths given by the user (with prefix -I) in the buildconfiguration->platform->compiler options with the help of an API string tokenizer. String Tokenizer tokenizes the compiler options but it is a deprecated API and the expected tokenizing operation failed. Due to which, all the paths included by the user in the compiler options were not sent as wildcards/arguments to the CBMC tool and so the analysis resulted in RUNTIME_ERROR.

Removed the deprecated API and tokenized the compiler options string w.r.t to whitespaces using string splitter and updated all the references of the method. 

Added test code and unit test for analysis of external c files which has external headers in different locations.

Regards,
Dinesh


danielratiu:
looks good to me

PR #1703, opened 2017-05-24
 changing the c standard front end option to adapt cbmc 5.7

DineshLMS:
Hi,

The argument to be supplied to the CBMC changed from *--std99* to *--c99* and from *--std11* to *--c11*

Updated the strings.

Regards,
Dinesh


danielratiu:
I have performed a small additional fix to enable the use of earlier CBMC versions

PR #1716, opened 2017-07-01
Chunks annotated with keep alive in generation should not be deleted

joao-silveira:
LowPrioStuff mapping script always deletes non implementation chunks even if keepAliveInGeneration is attached.

I've changed the script to not delete chunks annotated with keepAliveInGeneration.

Can you take a look?


coolya:
I this still valid?

joao-silveira:
Can be abandonned. I don't need it anymore.

PR #1718, opened 2017-07-13
Language name visible in the code-completion instead of alias name in exported Base concept

aswinsugumaran:
Hi,

In the implementation module, after inserting the **exported** from base concept when you code complete you can see the list of concepts with language name instead of alias name(refer the image attached). I have changed to return with the conceptalias name of the language. Please review the changes and merge with milestone

export_name_1

Regards,
Aswin Sugumaran


coolya:
I'm not sure what the use case for this is? Why would the user care about the declaring language of the type instead of the alias (which is will actually see after inserting the type)?

aswinsugumaran:
Hi @coolya  previously it was working like this(refer the image attached) ,  its not appropiate to list the language name in the implementation module.

export_name_1_previous

DomenikP:
I have modified the implementation for ExportedDummy (commit ID b343eea3ec9bf67c0b50c7dc52cc4b1790610633) to let the code completion look like the default menu that is shown when instantiating the concepts on the module content level (e.g. function or global variable declaration).

aswinsugumaran:
@DomenikP  After the new fix the menu looks like this, i think its little bit unreadable. @DineshLMS  your take on this ?
exportedbaseconcept

PR #1729, opened 2017-07-26
Column header of model consistency table is not shown properly and first column is unreadable.

VigneshCEG:
Hi,

After we apply the look and feel for model consistency, the column could not be expanded and content of the column header becomes unreadable.This is because when look and feel is applied, the width of the column is not enough for that particular font and other specifications.So i have increased the column width of first column in the table model consistency.Now the content of column header is shown and it is readable.Please review the changes and merge with milestone.


DineshLMS:
This is regarding the analysis-results window, right?

VigneshCEG:
Yes it is related to analysis-results window.

PR #1730, opened 2017-07-28
Fixed tutorial extraction and set to rename the setup.exe from the release to mbeddr-win-[date].exe in gradle

DorelComan:
Unfortunately there are mixed some nodes tracing tests, i realized that after pushing, but it should not be a problem


coolya:
Thanks for the pull requests. I'm quite busy at the moment with the 2017.2 migration. But I will have a look at it during this week.

DomenikP:
@DorelComan 
The changes in the tutorial and the Gradle script look good to me, but all of the test cases that you have modified/added fail on my local machine with a comparison erorr.

Can you please have a look at the tests.

danielratiu:
@DomenikP 
I am now having a look at tests, and I perform small improvements of the original code

PR #1740, opened 2017-08-16
WIP: MPS 2017.2 migration

coolya:


coolya:
The Model Listeners seems to be broken. The register method throws and `UnsupportedOperationException` sind the API is finally deprecated. Since we have no tests for it the code still compiled and no test failed.

PR #1745, opened 2017-08-18
Create README.txt for Ecore Wiki folder

krishnanm86:


coolya:
maybe it makes more sense to put it here as a sub page? 

https://github.com/mbeddr/mbeddr.core/wiki

PR #1755, opened 2017-09-07
No type system error thrown for mismatch of units in assert statements

VigneshCEG:
For example consider assert equals statement:
Example:
double val;
assert-equals(3) d == 4 s;

There is no type system error thrown here as annotated type and non annotated type cannot be compared.So i have added type system check for units in assert statements.Please consider the changes and merge with milestone.


DomenikP:
I understand the problem you are describing, but do not confirm with the implementation that you have provided as introduces a dependency from mbeddr.core to mbeddr.ext, which violates the mbeddr's dependency rules.

Please review your code and try to find a solution that does not require this dependency. I don't understand why it's not sufficient to just compare the types of both expressions. Is there a subtyping rule for units that causes this problem?

PR #1757, opened 2017-09-12
Milestone/v16 2

vsnarendran:
Updated the cbmc version to 5.8 in the Cprover Preferences Dialog
--


danielratiu:
@vsnarendran I have already implemented these fixes (and one more) - they should be now available

PR #1763, opened 2017-09-14
Changed the label name - context actions

vsnarendran:
Changed the label name Context actions (mbeddr) to context actions


sergej-koscejev:
FYI this name was changed *from* "Context Actions" to avoid conflicts with JetBrains' tool of the same name.

vsnarendran:
@sergej-koscejev We have removed the mps context-action plugin in the RCP, Now it wont create conflicts.

PR #1809, opened 2017-11-14
doc.spreadsheet: remove core.base dependency

coolya:


stephaneberle9:
Using INamedConcept instead of IIdentifierNamedConcept to get rid of the com.mbeddr.core.base dependency is perfectly fine with me.

arimer:
What about the addition constraints, which were inherited from the IIdentifierNamedConcept? Style and Workbook do not need them any longer?

coolya:
@arimer please check in the upstream RCP if this dependency change would fix the issues we had. Since I don't have the RCP here it's hard for me to verify.

arimer:
Hey, the problem is still there because there is a  dependency to `mbeddr.core.modules`
image

coolya:
There is also a dependency on the modules gen generator which can't get resolved. I will try to find a workaround for that as well.

stephaneberle9:
@arimer I don't think that the quite C-specific property contraint from IIdentifierNamedConcept is meaningful to workbook and style names in spreadsheets. Most probably, the IIdentifierNamedConcept has only been used out of habit when the workbook and stype concepts have been created but not because there is any real need for that.

coolya:
@arimer @joao-silveira can one of you please verify that these changes fix the problem in the upstream projects that only depend on the platform? 

If so please add a positive review and we can merge it.

arimer:
Picked the latest plugins from team city and tried it out. Error in the modules pool is gone for me.
See: 
image

PR #1816, opened 2017-11-21
inca: Introduced data flow analysis to taint impacted link accesses i…

krishnanm86:
Introduced an impact analysis that analyses the impact of an input of a latticeoperation and provides all the variables that are impacted by the input of the lattice operation (parameters)

This impact analysis in turn uses a visible variables analysis, which is also provided

This impact analysis is finally used by the Eval optimizer to cache the required interface parts that need to incrementalized.

Also, provided tests for the same.


szabta89:
This looks good to me, nice job!

PR #1828, opened 2017-12-05
CreateMbeddrRoots ActionGroup

arimer:
Currently it is not possible to contribute any other action group to `CreateMbeddrRoots`.

To allow this, we could create a nested action group as a content of CreateMbeddrRoots and add the update()-section there. 

Afterwards contribution to CreateMbeddrRoots (Add Roots) is possible.


arimer:
I'll check.

coolya:
Did you now fix the concern of @slisson? Maybe it would make sense not to do a commit amend when you address feedback on a pull request so we can actually see that something was done. 

We can still squash the commits when we merge the request.

PR #1831, opened 2017-12-13
Ignoring certain references while using AssertNodeEquals for node comparisons

Nandha1712:
When using AssertNodeEquals (com.mbeddr.mpsutil.compare) for comparing two nodes, currently there is no provision to ignore certain references alone during comparisons. So i have introduced a class IgnoreReferece(similar to IgnoreProperty class which exists already) and used it in the generator of AssertNodeEquals and AreEqualExpression.


Nandha1712:
I have updated this branch(Nandha1712:feature/ignoreReferences) with the changes you have mentioned. Can you review it and merge it to milestone/v16.1-1?

PR #1836, opened 2017-12-18
Interpreter: Migration for moving interpreters to a plugin solution

slisson:
With the upcoming MPS version interpreters are not allowed to be in the
same language as the concepts it references. They need to be moved into
a separate plugin solution. This is now automated.

There is a new entry in
MainMenu > Migration > Migrations > Interpreter: Move to pluginSolution

It will
- create a new plugin solution with the name
  "<language name>.interpreter"
- move all root nodes with concept from the interpreter language to the
  plugin model of the new solution
- add the solution to the build script in the same group as the
  language. The build script has to be visible in MPS (via a
  project/global library or opened in a second project in the same MPS
  instance).
- update references between interpreters
  (Interpreter.relationships.target) to point to the moved node.
- check if dependencies are migrated first.


slisson:
Before this migration can be used, the runtime needs a reimplementation of how it discovers interpreters.

PR #1844, opened 2017-12-21
doc: fix absolut path generation

coolya:
The looks up for that path difenetion didn't work correctly because it
was doing node comparison and they were from different (transient) models.

PathMapping lookup now done via name.


coolya:
Ok, merging this one an addressing the copy issue in a separate one.

PR #1848, opened 2018-01-10
core: IVisibleElementsProvider migation preparation

### LINKED ISSUE #1846: IVisibleElementProvider.visibleContentsOfType() should be virtual
Currently it is not possible to override this method: `visibleContentsOfType() `although it would make sense in some cases.

coolya:
Made `ListScope visibleContentsOfType(concept<> targetConcept)` virtual to allow
overwriting and also made the return type `Scope` instead of `ListScope`.

Deperecated the other methods in the API that take conceptNode paramters
and added overloads that use concep<> instead.

Changed the implementation of the conceptNode methods to call into the
new methods that use concept<>.

Reimplemented `findByName` with collection API rather than iterating the
sequence manually.

Added a comment explaning how a default implementation of `visibleContentsOfType(conceptNode<>)`
should look like given the fact that it is still abstract and needs to
be implemented but shouldn't duplicate the logic in `visibleContentsOfType(concept<>)`

Renamed the third overload that would return the `sequence` instead of the `Scope` to `visibleContentsOfTypeAsSequence`

Fixes #1846

### DIFF
--- a/code/languages/com.mbeddr.core/languages/com.mbeddr.core.base/languageModels/behavior.mps
+++ b/code/languages/com.mbeddr.core/languages/com.mbeddr.core.base/languageModels/behavior.mps
@@ -395,6 +395,12 @@
       <concept id="8465538089690331492" name="jetbrains.mps.baseLanguage.javadoc.structure.DeprecatedBlockDocTag" flags="ng" index="TZ5HI">
         <child id="2667874559098216723" name="text" index="3HnX3l" />
       </concept>
+      <concept id="2217234381367190443" name="jetbrains.mps.baseLanguage.javadoc.structure.SeeBlockDocTag" flags="ng" index="VUp57">
+        <child id="2217234381367190458" name="reference" index="VUp5m" />
+      </concept>
+      <concept id="2217234381367530195" name="jetbrains.mps.baseLanguage.javadoc.structure.MethodDocReference" flags="ng" index="VXe0Z">
+        <reference id="2217234381367530196" name="methodDeclaration" index="VXe0S" />
+      </concept>
       <concept id="8970989240999019143" name="jetbrains.mps.baseLanguage.javadoc.structure.TextCommentLinePart" flags="ng" index="1dT_AC">
         <property id="8970989240999019144" name="text" index="1dT_AB" />
       </concept>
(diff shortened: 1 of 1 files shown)


coolya:
I think that fact that we failed two times to call the correct methods that don't cause a SOE hints at the point that we need some more documentation on the purpose of the methods in the interface.

coolya:
If you choose to override the new method `Scope visibleContentsOfType(concept<>)` then you should use the default implementation proposed in the comment. If you don't choose todo so you simply implement your logic in the old style 'visibleContentsOfType' which simply returns the sequence. 

Every new implementation of a IVisibleElementProvider should override `Scope visibleContentsOfType(concept<>)` and choose to ship the default implementation for the old style method from the documentation. Maybe we should add this explicitly to the documentation to avoid confusion.

wsafonov:
Yes, but with the new implementation, if the new virtual method is not yet overwritten, calling any of these two public methods `visibleContentsOfType(concept<>)` or `visibleContentsOfTypeAsSequence(concept<>)` would lead to a recursion.

coolya:
Of course but there is no way around that I think because thats the whole idea behind the implementation in the comment, that I should be used when overwrite the other method. We might be able to provide a checking rule that checks for this cases and provides an error, but that would be fragile it could only really check for the exact implementation from the comment.

wsafonov:
Hmm, why don't we do it like I proposed and additionally instead of suggesting the clients to implement the old abstract method like this:
``` 
@Deprecated 
public virtual sequence<node<>> visibleContentsOfType(conceptNode<> targetConcept) {  
  visibleContentsOfTypeAsSequence(targetConcept.asConcept);  
}
```
suggest the following implementation until the mehod is removed:
``` 
@Deprecated 
public virtual sequence<node<>> visibleContentsOfType(conceptNode<> targetConcept) {  
  visibleContentsOfType(targetConcept.asConcept).getAvailableElements(null);
}
```

wsafonov:
Ok, forget my last comment - somehow I didn't get your last amended commit. Basically the method `visibleContentsOfType(concept<> targetConcept)` was already implemented right - we got back to its implementation again.

coolya:
Sorry my fault. I shouldn't use ammend and simply create new commit so that's obvious that I changed something.

arimer:
My previous concerns are solved.

coolya:
I added some more comments. I think it's good to get merged now.

PR #1859, opened 2018-01-25
New expressions for root mappings

arimer:
This feature introduces 3x new BaseLanguage expression which are only usable in the conditions of a root mapping rule:

**rootMappingHasBeenExecuted**
Can be used to check if a rootMapping rule has been already executed.

**setRootMappingHasBeenExecuted**
Can be used to flag a rootMapping rule as: hasBeenExecuted

**executeOnce**
Can be used to flag a rootMapping rule to be executed exactly once.

To ensure this logic the session object is used during generation. See <a href="http://127.0.0.1:63320/node?ref=r%3A817e5f99-00a7-4e0a-a330-9b858666e078%28com.mbeddr.mpsutil.blutil.rt.genUtil%29%2F4213334375078591265&project=com.mbeddr.mpsutil">here</a>

With these extension there is now no need to manipulate the session object yourself if you want to 
map a Node (which is not a root node) inside a rootmapping rule and you want to ensure that this mapping is done only once.


arimer:
Jup. Saw it. Need to check why.

arimer:
I've moved the expressions + generation implementation to some new language. `com.mbeddr.mpsutil.blutil.genutil` and created a new runtime solution. But I'm not sure if the name of the new language fits to the naming convention.

coolya:
This seems to cause some problems with next MPS 2018.1 ... because the generators use reflective queries. Have set anything in the generator settings manually regarding this?

arimer:
No. I just saw that @artem-tikhomirov commited some changes. But it was not clear to me what these changes are doing.

PR #1864, opened 2018-02-01
WIP: Fix for issue 1854. Fixing type system for power expressions that hav…

krishnanm86:
…e bases with annotated type

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.build/models/com/mbeddr/build/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.build/models/com/mbeddr/build/build.mps
@@ -2780,6 +2780,12 @@
             <ref role="3bR37D" to="90a9:PE3B26QCrP" resolve="org.apache.commons" />
           </node>
         </node>
+        <node concept="1SiIV0" id="7VbYB1xIuSX" role="3bR37C">
+          <node concept="3bR9La" id="7VbYB1xIuSY" role="1SiIV1">
+            <property role="3bR36h" value="false" />
+            <ref role="3bR37D" to="90a9:2NyZxKpUE9j" resolve="com.mbeddr.mpsutil.blutil" />
+          </node>
+        </node>
         <node concept="3rtmxn" id="3xFG3bj5chC" role="3bR31x">
           <node concept="3LXTmp" id="3xFG3bj5chD" role="3rtmxm">
             <node concept="3qWCbU" id="3xFG3bj5chE" role="3LXTna">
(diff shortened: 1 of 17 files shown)


krishnanm86:
Thanks for the feedback

The annotated type (and the information about the specifications which involve the units)  are in the AnnotatedType. So the way we do it, where we just set it to double will ignore the annotations(units info)

I can create an interface that lives outside the units language and have a behavior method which is implemented in the annotatedtype. That should remove this ugly dependency.

Regarding the test solution and the debug prints, I will handle them.

coolya:
I think an interface is not required but the type calculation should use operation types similar to other arithmetic operations in mbeddr. Then the units language can provide an overloaded operations container to calculate the type and units correctly. You can have a look at how it's done with the `BinaryExpression`s in mbeddr.

krishnanm86:
I gave it some thought and why cant the type of a powerexpression always be the type of the "base" of the power expression, with some check to ensure the type of the exponent is a double? Is there any scenario where the type of the result of  power expression will not be the type of the base expression. This way we dont need to have any fancy logic. We just lift all the required information from the base (be it the type or the units).

coolya:
There are two things here, the first is that the type of the power to expression is always double which is on the safe side but if all operants (exponent and base) are integers the result will be an integer. We might not easily infer the size so int64 would be the safe bet here.

The second thing is that we want to use the exponent in unit result. so we square a variable with unit meters we know that it's m²

krishnanm86:
The first thing, I agree with you. Its a safe assumption to make. 

The second thing about getting the type of a power of a variable that has physical unit requires the context of both the fact that we are calculating the type of a power expression (in the math language) and the physical unit itself (in the unit language). For this, I will need to come up with some clever way of introducing may be an adapter language (like with components and state machines) and marker interfaces. I don't see any other way around this. I cant simply override the type of annotated type as we need to construct a new type (with the unit multiplied with the exponent: UnitReference).

coolya:
We don't need a marker interface I think. We just need to use the operation type feature of the typesystem. That adapter language then simply contributes the required operations types. As it is done for other operations like multiplications and devisions like it was done here:

http://127.0.0.1:63320/node?ref=r%3Af5260afd-8327-4c3e-bf02-c81ea8a33729%28com.mbeddr.ext.units.typesystem%29%2F4959640877382136011&project=com.mbeddr.ext

In a similar rule you can implement the logic to calculate the resulting unit.

krishnanm86:
Good idea. Will look into this.

krishnanm86:
Changes made. There is currently a math-unit-adapter language that defines overloaded operations for power expression in the context of annotatedtypes.

coolya:
Something looks strange in the build scripts, why is the adapter language in the `com.mbeddr.ext.test` build script? That way I won't end up at the end users and is only used during tests. 

Second in the build script of mbeddr the math language still has a dependency on the units language, can you check why this is the case? I guess a "reload modules from disk" can fix it. 

Other than that everything looks fine.

Issue #1722, opened 2017-07-20
Facet for coping generated documentation


coolya:
Create a facet that copy every documentation related file into doc_gen folder next to the source_gen folder

joao-silveira:
More or less related, there was already a facet for copying in the generated images:

http://127.0.0.1:63320/node?ref=r%3Aa38f0e6e-eb07-4712-927b-30c9e021556c%28com.mbeddr.doc.gen_xhtml.plugin%29%2F205505736075693674

I guess it got broken somewhere in time.

coolya:
Interesting, didn't know about that one. 

What do you think makes more sense, trying to repair it or rewrite it with the experience of the past weeks?

alexanderpann:
This was implemented in #2240.

PR #1867, opened 2018-02-02
add MPS versions back to file names

coolya:
As @eugenschindler reported in #1722 due to our build infrastructure changes, the build number was no longer part of the file name that was uploaded to github. 

The feature has been added again. You can have a look on a release here:

https://github.com/mbeddr/mbeddr.core/releases/tag/nightly-557

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import de.itemis.mps.gradle.GitBasedVersioning
+
 ext.releaseRepository = 'https://projects.itemis.de/nexus/content/repositories/mbeddr'
 ext.snapshotRepository = 'https://projects.itemis.de/nexus/content/repositories/mbeddr_snapshots'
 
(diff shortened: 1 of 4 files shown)


eugenschindler:
Looks perfect! Thanks :)
Just a small question before merging: do you think it's a good idea to call it platform-distribution-MPS2017.3.2.zip instead of platform-distribution-2017.3.2.zip? This would make it more explicit that it's based on the MPS version. Or is there a good reason not to do that?

coolya:
Ho about including both, the MPS version and the mbeddr build number in the file name like `platform-distribution-1.0.19813.671048f-MPS-2017.3.2.zip`? That build number is also a valid version number to resolve mbeddr from our nexus.

eugenschindler:
If that is not too much work, that would definitely be the most ideal solution!

eugenschindler:
Super! Thanks a lot!!

eugenschindler:
Hi Kolja,
Unfortunately, we lost the MPS version again in the build. I see again that the allInOne zips don't have an MPS version. Would it be possible to bring it back? I have no idea right now which version of MPS to use with, for example, nighly 571 (currently the newest).

PR #1880, opened 2018-02-22
mpsutil: allow multi xml file import

coolya:
The xml importe action in mpsutil did not allow to select and import
multiple files. It's tedious to import a larger set of files. The
change to add this feature was pretty simple to implement by simply
itterating over multiple files.


arimer:
Looks good.  👍 
We should also think about allowing the import o a whole directory.

PR #1889, opened 2018-03-16
actionFilter: Implemented toobar customization

joao-silveira:
Now with the action filters language is also possible to configure the actions available in the toolbar.

Besides configuring the available actions it is also possible to customize icons.


slisson:
Instead of the hardcoded list CustomToolBar.defaultEntries, you could save the existing entries in IdeActions.GROUP_MAIN_TOOLBAR before you modify it.

joao-silveira:
You are right. fixed it.

PR #1890, opened 2018-03-16
Added functionality and tests to determine sinks in a graph and check…

AlexeiQ:
… whether the graph is connected. Did also a small refactoring for better understandability of the Graph methods


szabta89:
Looks good, a few comments:
`isConnected`

AlexeiQ:
I used the visited-hashmap, because the algorithm needs to check lots of times, whether a node is already queued and with a set I would need to search/find the value each time instead of accessing it directly by key, wouldn't I?

szabta89:
You should just add the elements to the set once it is visited. This is an O(1) operation. Then, asking the set if an element is contained (visited) is another O(1) operation.

szabta89:
This looks good to me. We can merge this once the build is fixed.

PR #1892, opened 2018-03-28
mpsutil: enhanced conceptdiagram with the possibility to add new languages in scope

danielratiu:
Now the concept-diagram can display concepts defined in other languages.


danielratiu:
Hi Joao *,*

I have extended the implementation to deal with AbstractLanguageIdentity. I
have also added an example in the sandbox language.

Please have a look again. Thank you!

Dan

joao-silveira:
Now it works, great!

I was also thinking about it, and following this ticket I had a couple more of ideas, see #1895 and #1896

PR #1894, opened 2018-03-29
smaller plugins in mpsutil

coolya:
Historically we had one big mpsutil plugin. Loading it was a all or nothing choice. I removed some of the bigger "subsystem" from the single plugin and created small plugins out of them. This allows fine grained loading in downstream projects. RCPs can easier decide which plugins they want to strip. 

It should also allow us an easier transition to mps-extenstion repo as we can move the plugins one by one. 

I also made sure that the big mpsutil plugin depends on the newly created ones. Downstream projects will not see any change until they decide to change their dependencies.

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -532,6 +532,36 @@
         <ref role="m$f5T" node="3quoVcnOFk5" resolve="group.suppresswarning" />
       </node>
     </node>
+    <node concept="m$_wf" id="64SK4bcNyQW" role="3989C9">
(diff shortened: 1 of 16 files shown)


sergej-koscejev:
What's up with the breadcrumbs-related changes? What is their purpose? I didn't understand that.

wsafonov:
We also need this restructuring of the mpsutil plugin the TV project. I've checked the new structure - looks good for me. Can we merge it to master? Based on it I would like to replace some coarse-grained dependencies in IETS3 plugins.

coolya:
@sergej-koscejev These breadcrumb changes where done because some interfaces in `core.base`that where giving introducing a dependency on breadcrumb for no obvious reasons. Most notably you could not contribute to the tree view for instance without implementing `IBreadcrumb`as well because the treeview interfaces would already extent them. 

In the documentation language I removed some implementations of the breadcrumb interfaces as well as none of the editors of the documentation language would show the breadcrumb selector anyway.

PR #1898, opened 2018-04-06
Added comment next to the to the "specify toobar" section

arimer:
Currently we will get a `ActionDuplicationException` when actions are removed from the profile and re-added to the **specify toobar** section. 
This commit just introduces an info that removing + re-adding the actions should not be done

### DIFF
--- a/code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.actionsfilter/languageModels/editor.mps
+++ b/code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.actionsfilter/languageModels/editor.mps
@@ -70,10 +70,12 @@
         <property id="1186403713874" name="color" index="Vb096" />
       </concept>
       <concept id="1186404549998" name="jetbrains.mps.lang.editor.structure.ForegroundColorStyleClassItem" flags="ln" index="VechU" />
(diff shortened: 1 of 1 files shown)


code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.actionsfilter/languageModels/editor.mps
```diff
@@ -439,11 +441,27 @@
...
+        </node>
+        <node concept="3F0ifn" id="7I52n9WCCwe" role="3EZMnx">
+          <property role="3F0ifm" value="specify toolbar" />
+          <node concept="VechU" id="7I52n9WCCwf" role="3F10Kt">
+            <property role="Vb096" value="DARK_BLUE" />
+          </node>
+        </node>
+        <node concept="3F0ifn" id="6hyuZyX9wzG" role="3EZMnx" />
+        <node concept="3F0ifn" id="7I52n9WCDhs" role="3EZMnx">
+          <property role="3F0ifm" value="Toolbar actions cannot be filtered out from the whole action profile" />
```
coolya: Can we change this to something like `Make sure actions from the toolbar aren't filtered elsewhere` and maybe also mention that it would cause problems at runtime otherwise?

arimer:
Changed comment to the proposed one.

coolya:
I guess the next step would be to create a checking rule that check for theses cases. I think we should put a ticket in the backlog.

arimer:
I've created an issue here: #1900

PR #1912, opened 2018-04-25
actionsfilter: don't load persisted profile by default

coolya:
If no persisted profile is stored by default profile is choosed the
logic would first apply the default profile and then still try to load
from the persisted state which then would fail.

We now onload the persisted state if there is a selected persited
profile.


joao-silveira:
@till-f you might also be interested in this PR.

PR #1936, opened 2018-06-25
Removal of graph stream languages in milestone/v17.0

Nandha1712:
In the milestone/v17.0, the languages of graphstream are not required. Hence removing them.


Nandha1712:
Since graphstream components(gs-algo,gs-core,gs-ui) are less maintained and also these components are a part of LGPL licenses 
which causes a overhead in OSS clearance,it would be good if you could remove these graphstream languages.

krishnanm86:
@Nandha1712 Why not just remove the reference from the build scripts/rcp to ensure this language is not part of the artifact produced by artifacts.

PR #1941, opened 2018-06-26
isApplicable functionality for run configuration

DineshLMS:
Hi,

We want to execute Run/Debug only if a condition applies. There is no "isApplicable" method in the run configuration. To achieve this functionality, I have introduced a new "before_task" as "isApplicable". Always the condition has to be checked before Run (or even before rebuilding the model). So the "isApplicable before_task" has to execute first. 

All the before_tasks are registered to the BeforeRunTaskProvider extension point. Before_task which is registered first to the extension point appears first during the execution. All the before_task root nodes of the model are collected and then registered to the extension point. During the collection of roots, "rebuildMainModel before task" always appears first because it was created first and present first in the tree. But we need "isApplicable before_task" to execute first before the "rebuildMainModel before task".

 In order to make the "isApplicable before_task" to come before "rebuildMainModel before task", I had to clone "rebuildMainModel before task" and delete the original one.  Cloning had created new before_task and assigned a new node id. So during the collection of root nodes, the "rebuildMainModel before task" appears after the "isApplicable before task". Thereby achieving the order.

Inside the "isApplicable before task", I am collecting all the extensions of the extension point and calling the "isApplicable method" of the extension object.

An extension point is created to allow ESD to get a hook where we can specify the condition after which the Run/Debug should get execute.

Please review and merge into milestone/v17.0-1

Regards,
Dinesh


krishnanm86:
@DineshLMS Everything looks fine, except that the "rebuildMainModel" node seems to show as deleted and re-introduced with the same contents as shown in the image attached. Could you please ensure this is not the case. 

image

DineshLMS:
@krishnanm86 re-introduction of rebuildMainModel is intentional. I have explained the reason in the above comments

PR #1943, opened 2018-06-30
C++ Support

nickdonnelly:
## In this PR

C++ Support including:
* Classes
* Inheritance
* Namespaces
* C++ -style Casting (eg `static_cast`, `reinterpret_cast`)
* Operator Overloading (experimental due to an MPS bug in building the typesystem)
* Templating
* Text Generation
* Various C++ keywords (eg `new`, `delete`, `constexpr`, etc.)

## Minor (non-breaking) Changes to mbeddr.core

Unfortunately, the implementation of certain concepts within mbeddr are specific to the C language, and cannot be extended easily. As a result, we have had to make some (small and most importantly, non-breaking) changes to mbeddr.core in order to make our code work. Below is an example to illustrate the kind of changes that we made: 

screenshot

We added a boolean property `isCPP` to the concept ImplementationModule in com.mbeddr.core.modules. The default value for this boolean is false, and it does not influence anything inside of mbeddr. However, in our C++ implementation module, we set this boolean to true, and if it is true we override the implementation module in order to prevent c-specific text generation from occurring (eg `extern C` blocks being placed in the header). This was hard-coded into mbeddr.core, but obviously had to be changed in the C++ implementation.

In addition, the `stdint.h`, `stddef.h` and `stdbool.h` imports for the C implementation module, previously hard-coded in its generator elements, have been moved to the behavioral function `importsForHeader` within the behavior of the C implementation module. This does not touch the external module, nor does it adjust the functionality of these header imports for any of the mbeddr C modules, as they are still hard-coded in a different way. This simply allows us to remove these header imports from the header files generated from the C++ implementation module, and replace them with the `cstddef` and `cstdint` when necessary.

While these commits do modify mbeddr.core, they do not influence the functionalities of mbeddr in any way. All commits that change mbeddr.core have been thoroughly checked, and the ones that were unnecessary or unintended have been reverted.


alexanderpann:
I am closing this PR because there was no interest in recent years to support C++ in this repository. It would also be a major effort for us to maintain this functionality. Even the C functionality is already little maintained.
The most recent C++ implementation by the same authors is https://github.com/DSLFoundry/mbeddr.cpp which supports MPS 2020.3.

PR #1944, opened 2018-07-03
Adding null check to ICOntrolledName in check_IControlledNamedConcept

Aarthi6574:
In com.mbeddr.core\languages\com.mbeddr.core.base\languageModels\typesystem a null check is missing in the check_ICOntrolledNamedConcept .Hence adding a null check for the IControlledName.


coolya:
For some reason the line ending in this file has changed therefore git things the entire file was changed. Can you check your git settings if the line ending is configured correctly?

krishnanm86:
@coolya I am having the same issue. The diff in MPS points out to the proper if statement, but git seems to think the whole file was re-introduced. Possibly because of some OS related stuff. 

@Aarthi6574 Please try to re-commit with the change inside the file and not the whole file. Other than that, the check itself looks fine to me. 

If needed, I can make the change directly to the milestone, since it is just one expression. Keep me posted if that would be required.

Aarthi6574:
Branch feature/bugfixes has been pushed as I have re-commited the changes once again which replaced equals() with :eq: to avoid NPE in check_IControlledNamedConcept.

PR #1962, opened 2018-07-17
Column header in analyses window is shrinked and is unreadable 

VigneshCEG:
Hi,

After we apply the look and feel for analyses tables (raw counter example and the results table), the column could not be expanded and content of the column header becomes unreadable.This is because when look and feel is applied, the width of the column is not enough for that particular font and other specifications.So i have added a method to set the max width with respect to table and font size .Now the content of column header is shown and it is readable.Please review the changes and merge with milestone.


VigneshCEG:
This issue is similar to https://github.com/mbeddr/mbeddr.core/pull/1729

PR #1976, opened 2018-08-02
show comprehensible error message if not building with java 8

abstraktor:
@coolya

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,9 @@ ext.artifactsDir = new File(rootDir, 'artifacts')
 ext.ciBuild = project.findProperty("forceCI") ?: !project.findProperty('mpsHomeDir') && project.findProperty("teamcity")
 
 buildscript {
+    if(JavaVersion.current() != JavaVersion.VERSION_1_8){
+        throw new GradleException("This build script requires java " + JavaVersion.VERSION_1_8 + ", but you are currently using " + JavaVersion.current())
+    }
     repositories {
         mavenLocal()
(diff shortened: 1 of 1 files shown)


sergej-koscejev:
I was requested to review this but I'm missing context. @abstraktor why is this needed? What happens without this check?

abstraktor:
Without the check, it fails with an error message close to „I can’t find a JDK in the folder: “ and then pointing me to my JAVA_HOME.

The confusing part is, that this path actually contains a JDK, but for java 10. So it was just not the version it was looking for. If you would like me to, I can reproduce it on monday and post the message here.

abstraktor:
For completeness sake, here the message without this patch:

```
* What went wrong:
A problem occurred evaluating project ':build:com.mbeddr'.
> Was not able to locate jdk home folder. Use 'jdk_home' project variable to specify JDK location explicitly. Current JAVA_HOME is: /Library/Java/JavaVirtualMachines/jdk-10.0.1.jdk/Contents/Home
```

coolya:
Sorry for being late to the party but had lot's of other stuff on my todo stack. 

I tried if you could get rid of the code that manually tries to locate the `tools.jar` when building with a JDK newer than JDK 8, as the docs claim that the classes from the `tools.jar` are now automatically available via the `ApplicationClassloader` but MPS completely fails when running on JDK 9+. It can't even build stubs for the JDK in these cases then. 

I think we should add similar code blocks to the other MPS based build scripts as well [(IETS3](https://github.com/IETS3/iets3.opensource), [MPS extensions](https://github.com/JetBrains/MPS-extensions), etc)

PR #1978, opened 2018-08-07
build:changed mps version to mps2018.2 final

joao-silveira:


coolya:
Can you sent a PR to the MPS-extensions as well?

PR #1999, opened 2018-10-12
Updation in Success Test Message of Unittest Helper

aswinsugumaran:
Hi All,

From the unit-test result message(**Unittest_Helper**), there are some chances(some outside simulation which uses TestResult.xml file might fail) that tests may not be completed. In those cases the result message in the console says  "Tests execution finished successfully!". This is somewhat confusing to user since those tests were never complete. So I have removed Successfully from the print statement and formatted some statements. 

Please review the changes and merge with milestone/v17.1-1

Kind Regards,
Aswin Sugumaran


krishnanm86:
@aswinsugumaran  Could you please squash the commits into one commit. It looks like there is only file that has been changed and yet there are 7 commits, which would make it hard to review and also ugly after merging. Thanks in advance.

PR #2008, opened 2018-11-07
Feature/asynccell

abstraktor:
see readme


wsafonov:
@slisson note that sequential computation might be a desired functionality because of the dependency between the values computed for the asynchronous cells (next to be computed value might depend on the result of a ongoing value computation or some of its intermediate results)

till-f:
@slisson @abstraktor 
Replacing the AsyncCellUIUpdater by a Timer seems to be a good idea (it is only responsible to give some feedback to the user by "rotating" the character).

It's some time ago, but IIRC the thread pool (IntelliJ) was used for all computation started by the AsyncChellManager. So maybe the AsyncChellManager is not responsible for stopping any threads. Because there is only one worker anyway (regardless of the number of async cells), I don't believe that it will make any difference to suspend it when there are no active cells.

slisson:
I didn't mean to literally  stop the thread, but to stop the task that you submit to the thread pool. The while loop never stops and will occupy the thread forever.

till-f:
Ok, that's true, but do you think the user will notice any difference (in performance)? At the moment the logic can be really simple, but I see some pitfalls ahead when we start/stop the task dynamically (e.g. cleanup after class (re-)loading).

slisson:
You need to exit your while(true) loop when the plugin is disposed. This is currently not happening.

abstraktor:
Summary of my changes:

- asynccell: replaced AsyncUIUpdater with javax.Timer
- asynccell: refactored AsyncCellUpdater to use IntelliJ thread pool and be timed by a javax.Timer
- asynccell: AsyncCellUpdater thus files every update cycle as separate task into the thread pool (parallel evaluation is a small change now, yet didn’t seem desired for now)
- asynccell: both tasks (spinner and updater) get suspended when all async cells have their values computed
- added com.mbeddr.mpsutil.blutil.test.waitfor test language extension for asynchronously testing asynccell
- moved all languages to their virtual folders
- created automated tests for all three languages
- extracted stub solution `com.mbeddr.mpsutil/org.mockito` out of `com.mbeddr.debugger/tests/test.debugger.core`
- created sandbox for parameterizedMenu and asynccell (the tests of waitfor should be enough demo)
- parameterizedMenu: added a `parameterizedMenu.sandbox.sandbox` solution to showcase the menu described in `parameterizedMenu.sandbox`
- added the three languages and stub solution to mbeddr.platform build script
- added the three test solutions to mbeddr.platform ci test script

These commits are not always semantic. So I'd recommend to review the whole diff.
If the review passes, we should consider squashing the commits before merging to keep the master clean. I preferred to keep the more detailed history for now in case of something went wonky somewhere.

slisson:
Looks better now. There is one mistake in the build script. You added the mockito jar to the actionsfilter plugin. It should go into a lib folder inside the mockito plugin.

abstraktor:
muted the tests and posted this issue to the mockito mailing list: 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/mockito/aK-m4CF7InM/q6M_XX8PAwAJ

PR #2011, opened 2018-11-09
Target within binaries is forced to Win32- not able to nullify

aswinsugumaran:
Hi All,

Follow the steps to reproduce the error,
1. Add a build configuration and select GNU in the platform section.
2. Add static binary in the binary section.
3. Remove the platform from the build configuration. 

From the image attached you can see the target reference in the static binary is broken.

image

So I have added a keymap for keys(delete,backspace) to the GNU platform editor to delete the referenced target from the binary section. 

Please review the changes and merge it to milestone/v17.1-1

Kind Regards,
Aswin Sugumaran


krishnanm86:
Hi,
   This is a bug that the user is unable to nullify the broken reference. But the proper way is not to associate a node detach for select keys. This solution is non-generic and editor level solution to a model level problem. I have pushed a fix where the target does not become invisible everytime the reference is broken. This will allow the user to handle the broken reference just like any other model.

Issue #1991, opened 2018-10-02
The "default image path" in DocumentConfig can not be referenced by an ImageParagraph.


kbirken:
Use case: Embed an image into documentation written with com.mbeddr.doc.

When using the ImageParagraph concept from com.mbeddr.doc, a path for an image file has to be specified (see "from" element). However, it is not possible to reference the path which has been specified as "default image path" in the DocumentConfig.

It is possible though to reference the "temp" path from the DocumentConfig. So currently images can only be provided by storing them into the file system folder referenced by the "temp" path.

alexanderpann:
It looks like it was fixed in #2016.

Issue #1992, opened 2018-10-02
Images from an ImageParagraph are not shown in HTML output.


kbirken:
Use case: Embedded an image file into documentation using ImageParagraph and show it as part of the generated html output.

During HTML generation, the image file should be copied to the images-folder inside the doc_gen tree and the reference inside the HTML code should have the proper path. This is how it currently looks like:

```
  <div class="imageWithBorder">
            <img src="doc_temp/CreateVariabilityChunk.png" />
</div>
```

Here, the doc_temp folder right below the solution root has been used as storage location for the image file. However, the generated HTML file is located deep inside the doc_gen folder and thus the relative path is not working.

alexanderpann:
Closing because it seems to be fixed in PR https://github.com/mbeddr/mbeddr.core/pull/2016.

PR #2016, opened 2018-11-21
com.mbeddr.doc: refactored image handling

joao-silveira:
Image handling was broken, namely:
-Default image path didn't show up in the scope of the path for resources (such as for ImageParagraph)
-When using ImageParagraph, the images would never end up in the doc_gen output folder and so we always had broken links

So I did the following refactorings:
-Now it is mandatory to configure the default image path
-Default image path folder is copied in the doc_gen output folder
-Screen shots also end up in the copied image path in the doc_gen folder
-Deprecated DefaultImagePath as this seemed a half baked idea. Provided a migration to replace it with a regular PathDefinition

From now on default image path in the documentation config must always be properly configured, and this resource should always be used as resource for ImageParagraphs.

Also I wanted to create a new facet, that would run after placeGeneratedFiles for copying the default image path folder into the doc_gen folder, but failed to to so. Therefore I included it in placeGeneratedFiles itself. If you feel like it is still a good idea to move it in a new facet we can try again.


joao-silveira:
This should fix: #1991 and #1992

joao-silveira:
I wasn't sure if I should:
1. Place the screen shots in the original image folder and then let it be copied with all the other files to the doc gen
2. Place them directly in the doc_gen folder

For now i've chosen the latest one, not sure though if this is what the user would expect.

PR #2024, opened 2018-12-07
build: fixed maintenance branch versioning

joao-silveira:
I am not sure if this is the optimal solution. However, should fix the problem that the artifacts built from maintenance version have the branch name as a  prefix. We need to fix this so that we can use the agreed "1.1.+" version specifier to get the latest platform.


coolya:
Change looks good but seems there is a problem with the build infrastructure at the moment. When that is sorted out we can run the build again.

joao-silveira:
All back to green. merging it.

PR #2028, opened 2018-12-12
NPE in IVisibleElementProvider.visibleContentsOfTypeAsSequence

abstraktor:
# Suggested Change

IVisibleElementProvider.visibleContentsOfTypeAsSequences no longer raises NPE if .visibleContentsOfType returns null

- visibleContentsOfType sometimes returns a null scope
- In such case, visibleContentsOfTypeAsSequence currently throws an NPE
- let's return an empty sequence instead to make it more consistent and fix the following background issue

# Background
- aNode.visibleContentsOfType(…).getAvailableElements(null) is commonly used.
- Let's allow refactoring such occasions into aNode.visibleContentsOfTypeAsSequence(…) to replace the NPE with an empty set of elements
- this case is tough to debug since mixing behavior and baseLanguage methods, while the first do raise NPE and the second don't raise NPEs when calling them on null objects
- the setting in which this exception was triggered from iets3.components was fixed as part of commit 8e437aae963 in iets3.opensource on 20.6.18, yet SecA doesn't contain that commit yet

### DIFF
--- a/code/languages/com.mbeddr.core/languages/com.mbeddr.core.base/languageModels/behavior.mps
+++ b/code/languages/com.mbeddr.core/languages/com.mbeddr.core.base/languageModels/behavior.mps
@@ -2,7 +2,7 @@
 <model ref="r:fd2980c8-676c-4b19-b524-18c70e02f8b7(com.mbeddr.core.base.behavior)">
   <persistence version="9" />
   <languages>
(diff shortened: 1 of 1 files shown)


coolya:
The change itself looks good to me. 

But two things:

- the API that allows access from the IVisibileElementsProvider that implicitly resolves a scope and returns the elements will be removed in the near future. In fact `visibleContentsOfTypeAsSequence` should be deprecated. 
- please don't stat the obvious in code comments: `// no NPE for null scopes`

PR #2036, opened 2019-01-24
c.m.m.e.QueryList: catch INSERT in front of a QueryList

abstraktor:
# Problem
Given I have a node which shows querylists as well as RefNodeLists
When I focus a constant cell outside these lists
And I hit RETURN
Then the next upcoming RefNodeList will receive an INSERT action with a null anchorNode
But unexpectedly, QueryLists will not be considered

See how it jumps over the QueryList:

# Rationale
- the reason seems to be that QueryLists return `null` when calling their `getSRole()`
- `ChildrenCollectionFinder#find` seeks for the next cell with `EditorCell#getSRole().isMultiple() == true`
- `EditorActionUtils#callAction` uses it to find the corresponding action and run it

# Approach
- I made the `EditorCell_QueryList` set a dummy `SAbstractLink` as SRole of itself which always returns `true` on calls to`isMultiple()`. The other methods return some null objects.
- Then I needed to fix `QueryListHandler#getAnchorNode` to return null when being provided with cells outside query list cell. So far the INSERTs only came from inside the cell, but now the anchorNode may be null.

As a result, the issue is fixed:

# Todos

- [x] a querylist cell now provides getSRole().isMultiple() == true so that actions propagate similar to RefNodeLists
- [x] manual test case [RootWithChildren](http://127.0.0.1:63320/node?ref=c58b797b-c53c-4442-abff-4d54c72169ea%2Fr%3A5f236bd8-eb21-40b6-825a-9e614ca27175%28com.mbeddr.mpsutil.editor.querylist.sandbox%2Fcom.mbeddr.mpsutil.editor.querylist.sandbox%29%2F1952741127689876949&project=com.mbeddr.mpsutil) added in the demolang and sandbox
- [ ] this change may require existing QueryLists with _insert new_ concept functions to deal with null `anchorNode`s. If it is null, `insertBefore` should not be considered anymore and it should be prepended. Maybe we should catch this in the cell and always set `insertBefore=true`? It would still be an INSERT action though.

### DIFF
--- a/code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.editor.querylist.demolang/languageModels/editor.mps
+++ b/code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.editor.querylist.demolang/languageModels/editor.mps
@@ -3,7 +3,8 @@
   <persistence version="9" />
   <languages>
     <use id="120e1c9d-4e27-4478-b2af-b2c3bd3850b0" name="com.mbeddr.mpsutil.editor.querylist" version="-1" />
-    <use id="18bc6592-03a6-4e29-a83a-7ff23bde13ba" name="jetbrains.mps.lang.editor" version="12" />
+    <use id="18bc6592-03a6-4e29-a83a-7ff23bde13ba" name="jetbrains.mps.lang.editor" version="-1" />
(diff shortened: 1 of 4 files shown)


coolya:
@wsafonov I think you used query list quite a few times. Can you comment on the API change required for this?

slisson:
anchorNode was never guaranteed not to be null. In case of an empty querylist, `insert new` was always called with `null` for `anchorNode`. Your change isn't an API change.

PR #2043, opened 2019-01-31
Fix Java compile problem in Interpreter generator

kbirken:
Building an interpreter for a concept which has a backslash character in its alias will lead to a Java compile error. This is because the backslash will be part of a string in the generated code and will be interpreted as an invalid escape character. This PR fixes this by duplicating each backslash in the corresponding Java string.


szarnekow:
I cannot judge this from the xml diff, but do we also need to escape double quote chars? Or are these already handled somewhere else?

kbirken:
Good point, " would probably also break the compatibility. I will test this and fix it, if necessary.

PR #2065, opened 2019-04-14
filepicker: fixed SolutionPathUtil hack

abstraktor:
# Problem
`com.mbeddr.doc.gen_xhtml` generates broken links to its image paragraphs when including documents from other models.

# Investigation
`SolutionRelativeDirPicker#getCanonicalPath()` sometimes returns relative paths, sometimes absolute paths. This leads to downstream errors when compiling a documentation with `gen_xhtml`, because it uses this path inside `UnixPath#relativize`.

The reason seems to be that `SolutionRelativeDirPicker#getPrefix()` is null in these cases, because `SolutionPathUtil.getSolutionRootPathForNode` returned null.

# Fix
The hack in `SolutionPathUtil.getSolutionRootPathForNode` seems to have been dead code because transient models never know their repository in 18.3 (as far as I could see). To carry on with this hotfix, the implementation know uses the repository of the target model.

```
module.myComponent.getRepository()
```
grafik

# Open Questions
- This fix assumes "The solution lives in the repository that we're generating into". Does it hold?
-  Whats the longterm way to do this? The hack seems needed when accessing the solution during generation.


abstraktor:
push, no interest in this? I consider this relevant for master as this is tough to debug

slisson:
- There is a duplicate of this class in `com.mbeddr.core.base.behavior`. We need to fix both or even better, remove the duplication.
- Why `isAssignableFrom` instead of instanceof?

abstraktor:
instanceof returns false for instances of subclasses. It thus breaks the [Liskov Substitution Principle|https://en.m.wikipedia.org/wiki/Liskov_substitution_principle] and I thus consider it a smell to use it without dedicated need. isAssignableFrom works covariant as desired in polymorphic programming.

slisson:
> instanceof returns false for instances of subclasses
Were did you learn that? That's not true.

abstraktor:
*after googling and scratching my head...*

True. It seems to only relevant when not knowing the type at compile time, so not too relevant here. I’m ok with changing it.

PR #2071, opened 2019-05-24
Visitor pattern implementation proposal in order to customize default MPS persistence.

kather-siemens:
Visitor pattern implementation proposal in order to customize default MPS persistence.


krishnanm86:
Please try out the MPS custom persistence and adapting it. It has ways to specifying concept kinds and nodes, and specifying how these need to be generated.

https://www.jetbrains.com/help/mps/custom-persistence-cookbook.html

kather-siemens:
While we were analyzing the possibility for overriding MPS default persistence, we found this cookbook.  It describes how to override the default persistence, and explains the methods offered by custom persistence API.

By following this cookbook, we can take control over the MPS persistence path by overriding save and load methods. Inside the overridden method, we need to implement how to persist the concept hierarchy from Implementation module to Blocks, ports, etc. We have chosen visitor pattern to persist this hierarchical information.

In visitor implementation, it is essential to start the visit from root node of model to all its children. Hence we need some hook to trigger the visit from Implementation module as well.

PR #2076, opened 2019-05-29
Interpreter: added possibility to query an env without consulting the parent env

bkolb:
Cherry-picked from maintenance branch


bkolb:
the tests failing seem to be unrelated to me

PR #2093, opened 2019-08-19
Local variables inside multiplatform statements are not refactored as per C90 standards.

aswinsugumaran:
Hello All,

We have a local variable inside the multiplatform statement injected by the portability script.
C90 Scripts which refactors the local variable initialization takes care of only statement concept and not the **multiplatform statement** concept. So I have introduced another similar logic to refactor the local variable initialization statement contained in the multi-platform statement.

Please review the changes and merge with milestone/v19.2_20183

Kind Regards,
Aswin


aswinsugumaran:
@krishnanm86  We had a call on this above issue.

stephaneberle9:
The suggested changes of the c90Code preprocessing script contain the following issues:
- the logic for extracting local variable declarations and moving them at the top of the surrounding blocks is unnecessarily repeated
- the injected new MultiPlatformStatements contain hard-coded win32/othersHeaders
- the injected new MultiPlatformStatements only work under Microsoft(R) Windows(R) but not on other platforms

PR #2098, opened 2019-09-06
Macro value returned null in the macro file and folder picker

VigneshCEG:
When the 'getValue()' method inside the 'PathMacros' class ,is called by creating a non-static
object ,value of the macro is returned as null.So,reverted the instance creation to static again using
the getInstance() method in the 'PathMacros' class.

Please review the changes and merge with milestone.

Regards ,

Vignesh M

### DIFF
--- a/code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.filepicker/models/behavior.mps
+++ b/code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.filepicker/models/behavior.mps
@@ -2198,19 +2198,6 @@
       <ref role="13i0hy" node="5lKnBeAIfOh" resolve="getPrefix" />
       <node concept="3Tm1VV" id="7c_RIoB8IKt" role="1B3o_S" />
       <node concept="3clFbS" id="7c_RIoB8IKw" role="3clF47">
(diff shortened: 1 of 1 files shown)


krishnanm86:
@VigneshCEG The change currently reverts to the usage of Pathmacros.getInstance() which is deprecated since 3.4 and could be removed anytime. 

Just because the non static object's getValue() returns null does not mean it is the wrong solution. Replacing the Instance with constructor and then reverting again sounds like replacing one hack after another. 

The instance points to "this" and has no difference between the consructor and the getInstance itself. May be you need to call the init() after the getValue(). Please look at the source code :https://github.com/JetBrains/MPS/blob/master/core/project/source/jetbrains/mps/project/PathMacros.java

Please provide a robust solution to this by considering how the PathMacros works and if the recommended choice does not work as expected, we need to create an issue with Jetbrains instead of reverting back to a deprecated method(one bad coding practice) to avoid nullpointerexception(another bad coding practice).

VigneshCEG:
The issue is more than just a null pointer exception ,the macro picker functionality is completely broken ,for example if : $(macro)location is given , since the macro is returned null ,wrong path is taken , so it affects majorly in all the places in which path is given using macros.There might be a problem in jetbrains , but at the moment this might be a temporary solution to have (reverting to get instance) as the macro picker functionality is completely broken.

PR #2102, opened 2019-09-18
Workaround for MPS-30302

sergej-koscejev:
`node.getPresentation()` throws a NPE if `node` has a null `concept` (this may happen when the concept hasn't been compiled or could not be loaded). This causes problems in the merge dialog.

As a workaround check for node.concept being non-null before calling `getPresentation()`.


sergej-koscejev:
Rebased and merged directly to master.

PR #2111, opened 2019-12-12
interpreter: deprecating applicable languages

wsafonov:
**Applicaable Languages** section of the interpreter definition currently contains references to structure aspect models with concepts evaluated in the interpreter. Its only purpose is to restrict the scope of visible concepts in the interpreter.

I suggest to completely remove these section and use the default scope instead (all referencable concepts from imported structure aspects). This should not really affect editing experience, as selecting concepts for evaluators is not performed very often and seeing couple more concepts in the CC menu shouldn't bother. But it will help us to get rid of the old, deprecated `ModelReferenceExpression` used in `ApplicableLanguage` and the need to call `SModelrepository.getInstance()`, which is also deprecated and would be removed in one of the next MPS versions.

Instead I suggest to provide in the editor a read-only list of languages, evaluated in the interpreter, based on the concepts used in its evaluators. This would actually play a meaningful role, as it would allow to quickly check, concepts of which languages are evaluated in the current interpreter. This would especially come handy in big interpreters with a long list of evaluators, where several languages might be evaluated at the same time:

image


wsafonov:
> `ConceptTypeExpression_Constraints`:`it.getAllSuperConcepts(true).contains(concept/IType/.asNode)` can be simplified to `it.isSubconceptOf(concept/IType/.asNode)`.
>
Good point, I must have overlooked it.

> Also, would it be a good idea to remove the deprecated reference presentation code while we're touching this part?
> 
Sure, I was going to do that later in your branch, but let's fix it right away.

> `Interpreter_Constraints`: do we need to calculate and store languages as a property? How about simply calculating it in the editor in a model access cell?
Actually it's not stored, just calculated property. That was just the quickiest and simpliest way came to my mind to render multiline text. But probably I can refactor it to use custom cell provider returning multiline cell.

PR #2116, opened 2020-02-11
Replace deprecated static field reference with a static method.

fisakov:
The field MPSInternalPortManager.PORT has long been deprecated
and is removed in 2020.1.
getCurrentPort() is to be used instead.


sergej-koscejev:
mbeddr master is now using MPS 2019.3, should this patch have been merged to master?

fisakov:
@sergej-koscejev feel free to merge if you think it's safe. In MPS buildchain the branch mps/2019.3 is used.

sergej-koscejev:
Why don't you use master? Otherwise you're testing MPS with some hypothetical version of mbeddr that's not going to be released.

PR #2117, opened 2020-02-11
Fix reference to IdeaPluginDescriptor.getPath() which has been moved …

fisakov:
…to superinterface.

The latest IDEA platform's PluginDescriptor has all the methods from IdeaPluginDescriptor.


ashatalin:
The patch is not backward-compatible with 19.2

PR #2124, opened 2020-03-16
Replaced the algorithm that apparently did not work with a simple one…

markusvoelter:
…, that works.


markusvoelter:
I removed the prints
The trailing comma didn't matter, but I removed it

PR #2125, opened 2020-03-20
interpreter: speed up evaluator lookup

sergej-koscejev:
Store evaluators in a hash(multi)map by concept.


sergej-koscejev:
I have to fix it so that when there's no evaluator for a concept then evaluators for the superconcept are considered.

PR #2131, opened 2020-04-14
Fixes for "zb" customer project (from branch zb/19.3)

till-f:
This pull request contains 4 fixes that where required after migrating a customer project from mps 2018.1 -> 2019.3. Changes should not have implications for others but please review carefully.

### DIFF
--- a/code/languages/com.mbeddr.core/languages/com.mbeddr.core.modules.gen/generator/template/main@generator.mps
+++ b/code/languages/com.mbeddr.core/languages/com.mbeddr.core.modules.gen/generator/template/main@generator.mps
@@ -5038,6 +5038,44 @@
                 </node>
               </node>
             </node>
+            <node concept="3clFbF" id="30ZiFaB3bdE" role="3cqZAp">
+              <node concept="2OqwBi" id="30ZiFaB3ejg" role="3clFbG">
+                <node concept="2OqwBi" id="30ZiFaB3bJT" role="2Oq$k0">
+                  <node concept="3l3mFP" id="30ZiFaB3bdC" role="2Oq$k0" />
+                  <node concept="3Tsc0h" id="30ZiFaB3cXi" role="2OqNvi">
+                    <ref role="3TtcxE" to="tpck:4uZwTti3__2" resolve="smodelAttribute" />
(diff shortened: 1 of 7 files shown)


wsafonov:
Looks good to me. Just minor stuff:
- in `reduce_PlaceholderTextList_ListHandler` the API has really changed, so it doesn't make sense to make a `super.createNodeCell(elementNode)` call, cause it always returns null. Instead you can directly call `getUpdateSession().updateChildNodeCell(node)`
- in `DescriptorCollector`, I wonder if you can simply use `MPSModuleRepository.getInstance()` instead of extracting the repo from the first project. Because it's also a `SRepository` and we anyway need `MPSModuleRepository` for interating through modules, this might be a better match here. But you should better double-check this for your scenario (when welcome screen is shown).

till-f:
@wsafonov I wanted to be as safe as possible :-) But your suggestions are fine. I'll try it out and ping you again.

till-f:
@wsafonov It works fine for me with your suggested changes. But I discovered another somewhat related issue. I've fixed it in a separate commit:

`ActionTreeNode::isChecked()` passed `null` to `ActionManager.getId(...)` when the respective settings page is opened (File -> Settings) here:
http://127.0.0.1:63320/node?ref=f%3Adiff_new%23r%3A7171fd48-62d6-4c67-ab22-d7d6b8fa4653%28com.mbeddr.mpsutil.actionsfilter.runtime%40new%29%2F6942859170521566684

This causes an Exception because the argument is marked `@NotNull`. IMHO it's not intended to construct an `ActionTreeNode` with myAction==null, thus I prevented the cause for my case. This probably does not cover all possible cases, but again, I didn't want to change more than absolutely necessary.

till-f:
@wsafonov: update: please do not merge, there's another issue now (missing actions in the toolbar). so I guess that the last hotfix is problematic!

till-f:
Hm, no, everything seems to be fine. It must have been a side-effect of the now-fixed exception when I have clicked through the settings dialog to provocate it. I've removed the workspace.xml and config in the home directory and now all actions are back in the toolbar... please review and merge :-)

PR #2134, opened 2020-05-19
editingGuide: drop dependency on com.mbeddr.core

sergej-koscejev:
editingGuide pulled in IVisibleElementProvider. This dependency is now replaced with a custom interface, INodeExporter.


abstraktor:
looks good to me. Couldn't test it though. Next time, we maybe should create a sandbox language to showcase the scope-restriction.

PR #2138, opened 2020-06-19
Change caption text of a model-as-image in presentation mode.

kbirken:
The concept `ModelContentAsImageParagraph` from mbeddr.doc allows to embed model nodes (by reference) in documentation. A caption/description can be added to the model content. However, in presentation mode this description is not displayed, but a generic text like `Code from ...`. 

This PR ensures that the actual description is shown instead of the generic text (if in presentation mode). 

Note: In the old implementation, Ctrl-Click on the description text jumped to the embedded model node. I couldn't do this; so now the figure number is a navigatable node. If the reviewer has a hint how to make the description navigatable again, please comment.

### DIFF
--- a/code/languages/com.mbeddr.doc/languages/com.mbeddr.doc/languageModels/editor.mps
+++ b/code/languages/com.mbeddr.doc/languages/com.mbeddr.doc/languageModels/editor.mps
@@ -615,7 +615,7 @@
       </concept>
     </language>
     <language id="120e1c9d-4e27-4478-b2af-b2c3bd3850b0" name="com.mbeddr.mpsutil.editor.querylist">
-      <concept id="6202678563380238499" name="com.mbeddr.mpsutil.editor.querylist.structure.Function_GetElements" flags="ng" index="s8sZD" />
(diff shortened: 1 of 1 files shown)


wsafonov (approved):
The change itself looks valid. Regarding your question, you can cerntainly apply `hyperlink-node` style to the description and the hyperlink should work, but it would not look exactly like as one would expect - because single words in the text block are located in separate cells, each word would be underlined separately and not the complete text block when hoovering, as with URL hyperlinks in the browser. Probably the only way to support proper behaviour including multiline description would be using JLabel with a custom click handling.

kbirken:
> The change itself looks valid. Regarding your question, you can cerntainly apply `hyperlink-node` style to the description and the hyperlink should work, but it would not look exactly like as one would expect - because single words in the text block are located in separate cells, each word would be underlined separately and not the complete text block when hoovering, as with URL hyperlinks in the browser. Probably the only way to support proper behaviour including multiline description would be using JLabel with a custom click handling.

Thanks. I guess it's fine if the figure number is the clickable part for now.

Issue #2139, opened 2020-07-22
CombinedInterpreter sorting is not stable


abstraktor:
Given I have an interpreter `Int1A` in category `category1`
And I have an interpreter `Int1B` in category `category1`
And I have an interpreter `Int2` in category `category2`
And all of them have an evaluator for concept `MyConcept`

When run `new CombinedInterpreter(InterpreterEvaluationHelper.getInterpreter("category1"), InterpreterEvaluationHelper.getInterpreter("category2")).listEvaluators()`
Then I sometimes get `[Int1A.MyConcept, Int1B.MyConcept, Int2.MyConcept]`
And, **surprisingly**, I sometimes get `[Int2.MyConcept, Int1A.MyConcept, Int1B.MyConcept]`

# Analysis
* `InterpreterEvaluationHelper.getInterpreter` returns a `CombinedInterpreter`
* as a result, the above-written is comparing two `CombinedInterpreter`s, which don't have any relationship
* it seems that `InterpreterClassSorter.sort` is not a stable sorting algorithm (http://127.0.0.1:63320/node?ref=r%3Aea6cf71d-29d2-478d-8027-a9f4a4de53c4%28com.mbeddr.mpsutil.interpreter.rt%29%2F2447795128874040475)

# Severity

I need to combine my interpreters at runtime. Sometimes, all three should run, sometimes, only the category 1 should run. With this sorting working fine, I could rely on the interpreter caching instead of having to lookup the executable myself…

coolya:
You can always use "runs before" and "runs after" dependencies in case you have a specific order requirement, which sorting of the evaluators then takes into account.

abstraktor:
Yes, that's the sorting which is performed by `InterpreterClassSorter.sort`. However, these rules only express dependencies between usual interpreters, not between`CombinedIntepreter`s.

coolya:
Yes because we only support dependencies within a single category of interpreters which is represented as one combined interpreter. I would ague you use case of trying to combine different categories of interpreters and then have cross category sorting is rather exotic and not something we ever intended as a use case. It looks to me more or less like missing documentation and we could add a error message if an interpreter has a dependency to a generator from a different category to indicate that it will have no effect, would this have helped you to figure out the problem earlier?

abstraktor:
There is such a warning. That's why I'm fine with keeping Issue 2 asside. I suggest to only fixing Issue 1 for now.

[...]

PR #2140, opened 2020-07-23
mpsutil.common.graph topological is stable

abstraktor:
sorting is maintaining order of equal items (a.k.a. "a stable sorting algorithm")

this would solve Issue 1 of #2139

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,3 +8,26 @@ mbeddr aims at creating a different way of developing embedded software systems.
 At this point we are well ahead in developing an implementation of C in MPS which can then be used as a basis for domain-specific extensions. The development progress can be see from our [blog page](http://mbeddr.com/blog/).
 
 For more details please visit http://mbeddr.com
+
+
+# Contributing to the mbeddr.platform
+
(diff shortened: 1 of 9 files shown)


coolya:
Before we merge this, apart from fixing the obvious build problem, I will need collect some real world performance numbers on how this impacts performance, especially for interpreters.

abstraktor:
sorting the interpreters seems to be the only usage of common.graph from mpsutils. I asked around and didn't hear about any other project using it.

Regarding the perormance of sorting the interpreters: Our largest interpreter is the javainterpreter with around 10 interpreter nodes. So we're talking about performance of sorting 10 items, which shouldn't be of too high impact, I guess. Or are there places where the interpreter is called in such a frequent fashion that this turns relevant?

coolya:
The java interpreter is a toy example. That's why we need metrics from real world and yes there are places especially when collecting coverage information where hundreds of interpreters are called rapidly.

abstraktor:
Oh I see. I didn't know that.

coolya:
While the sorted version seems slower, most interpreter usecases will anyway use `InterpreterEvaluationHelper` which caches the interpreter for a given category hence the sorted evaluators are also cached. From an interpreter point of view I see now reason not to merge this change since the performance impact is not really relevant.

alexanderpann:
I am closing this PR because the changes were merged but not through this PR.

PR #2146, opened 2020-10-09
mpstutil.smodule: fixed GlobalModelAccess usage in addModelOperation

joao-silveira:
Because now GlobalModelAccess has no way to execute a command, we have to provide a project repository to the add model operation.

For now I made the repository optional so that we just don't break the existing usages. When a repository is not provided we try to get it from the opened projects.

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -7355,6 +7355,11 @@
             </node>
           </node>
         </node>
+        <node concept="1SiIV0" id="2rFgM0FWDQ7" role="3bR37C">
+          <node concept="3bR9La" id="2rFgM0FWDQ8" role="1SiIV1">
+            <ref role="3bR37D" to="ffeo:mXGwHwhVPj" resolve="JDK" />
+          </node>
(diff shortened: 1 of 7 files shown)


joao-silveira:
When we get it merged I will take care to port it to master.

arimer (approved):
Changes look fine to me. 

Just one minor remark. You introduced a model dependency 
`742f6602-5a2f-4313-aa6e-ae1cd4ffdc61/java:jetbrains.mps.ide.project(MPS.Platform/)`
which is not used as far as I see. Do you mind removing it before merging?

Just out of curiosity, how did you find out that `module.getRepository()` will not give you the required repo to run the command?

joao-silveira:
> Just out of curiosity, how did you find out that `module.getRepository()` will not give you the required repo to run the command?

Because it threw an UnsupportedOperationException.

coolya (approved):
I don't like the project manager hack a lot but I guess its fine for now.

I cleaned up the code and removed some deprecated classes from it which changed the method signature because arguments aren't required anymore. I suggest you take a look before merging again.

joao-silveira:
I see your point. I was trying to be the nice guy and provide a smooth migration path for the existing code. I assumed for most cases the "hack" should work and for others were it doesn't I don't think the consequences are too severe (Undo is not available?). The alternative I see would be throwing an exception and forcing all existing usages to provide the SRepository.

PR #2147, opened 2020-10-12
Fix ConcurrentModificationException in Documentation Plugin occurring on multiple editor components disposal

alex-y-wd:
A `ConcurrentModificationException` is thrown when multiple editor components are disposed (for example, when `Close All` menu is invoked or when project is closed).


szabta89:
Nit: why not just iterate over the `keySet` in `removeAllOnShowDocTabListeners` with a `foreach` instead of doing that through an iterator?

szabta89:
Looks good to me.

szabta89 (approved):

alex-y-wd:
> Nit: why not just iterate over the `keySet` in `removeAllOnShowDocTabListeners` with a `foreach` instead of doing that through an iterator?

Because the iterator is fail-fast and will throw `ConcurrentModificationException` if removal happens in the collection directly.

PR #2148, opened 2020-10-21
mpsutil.smodule: fixed generator

joao-silveira:
StorageType is no longer required.


arimer (approved):
Changes look fine to me. Thx for deprecating the outdated stuff. 👍

PR #2151, opened 2020-11-17
make some behavior methods virtual 

coolya:
which was requested by the CPP implementation based on mbeddr at https://github.com/DSLFoundry/mbeddr.core/blob/mps-2020.1/documentation/cpp/migration-issues.md


enikao:
Looks good to me.

PR #2152, opened 2020-11-30
mpsutil.smodule: fix storage type detection

coolya:
Don't ignore the passed in storage type and provide binary compatible
methods for existing code.


joao-silveira (approved):

joao-silveira:
@coolya I did two more changes:

* We were missing passing the storage type node to the switch. This way the switch would be ignored and fall back always to file per model.

* Undeprecated storage type child

PR #2154, opened 2020-12-08
Update to MPS 2020.2

coolya:


wsafonov:
MPS 2020.2.3 comes with a minor change in modules validity check during loading of modules that caused issues with analyses tests (see commit 3b4432c description). This change might also cause issues in downstream projects that have solutions with IDEA plugin facet. In situations where any of such solutions has to be correctly loaded, one would need to ensure that the corresponding IDEA plugin can be also loaded (e.g. using additional plugins specification in build model test configuration).

PR #2162, opened 2021-02-05
doc: fix code removal in screenshots

coolya:
Screenshots didn't respect the content removal specified in the
ModelContentAsScreenShotParagraph since we migrated to the logic
that takes screenshots from the intermediate model and not from
the original model.

This changes sets the editors hints for the intermediate model
correctly in the docs languages to allow the remove to work.

Only setting the hints wasn't enough because conditional editors
were disabled in the screenshooter code since the editor hints
for conditional editors were not set. The screenshooter now
always sets the conditional editor hint so that conditional

@slisson can you let me know if [setting](http://127.0.0.1:63320/node?ref=r%3A0a882e21-5553-485b-8777-3b0ace5a0d84%28com.mbeddr.core.base.pluginSolution.plugin%29%2F1344466372819333811) both conditional editor hints is required or not? 

image

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -18455,6 +18455,11 @@
             </node>
           </node>
         </node>
+        <node concept="1SiIV0" id="1aCwgFOnMPO" role="3bR37C">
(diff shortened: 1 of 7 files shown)


joao-silveira:
I'm not sure, but my understanding is that you need that hint for the case when you have multiple condition editors for the same concept. So I would say you need it.

In any case @slisson should confirm this :D

joao-silveira (approved):

slisson:
Yes, it's correct to set both hints.

PR #2164, opened 2021-02-15
doc: add support for multiple doc includes

coolya:
Remove the limitation that a document could only get included once. It
is now possible to include a document in as many places as desired. The
resolving of transient imports is supported and includes are resolved in
inverse topological order (from the leaf to the root). The generator
automatically restores references within the including document to
content that is included if required.

The reference resolving logic is limited to resolve references to
included content only to the first include. This means including the
same document twice in one document will not yield the desired result.

Please see the code comments for much more details on the design.


joao-silveira (approved):

joao-silveira:
Looks sharp 🦈

PR #2166, opened 2021-02-26
Forward-Port YSA Changes to mbeddr.core

abstraktor:
As discussed, these are the condensed changes we performed in YSA. We're now done with the migration. After merging this, we'd consider ourselves as "in sync" with common mbeddr mpsutils, knowing well that we might need to forward-port these changes when updating our used mbeddr version.

The commits summarize the activities well, I think


sergej-koscejev:
I don't like the idea of doing changes in 2019.3 first, then forward-porting:
1. I don't think we should spend effort maintaining 2019.3 anymore.
2. It's strange if 2019.3 gets extra features while 2020.3 doesn't.

Another thing, let's do one PR per logical change, otherwise it'll become a mess if changes are requested.

I'd welcome it if you could resubmit the PRs for master first, then we could backport changes to a few older versions such as 2020.2 or maybe 2020.1.

PR #2169, opened 2021-04-16
Permit any JSON value as root of a JsonFile instead of only JSON Objects

jonaskraemer:
The previous implementation enforces that a JsonFile contains a JsonObject as the root node of the contained JSON text.
This is not in accordance with the JSON Standard, which allows _any_ JSON value (such as arrays or a numbers in addition to objects) to be the root of a JSON text.
Refer to [https://www.ecma-international.org/publications-and-standards/standards/ecma-404/](https://www.ecma-international.org/publications-and-standards/standards/ecma-404/), Section 4.

Since a JSON object is a JSON value, this change does not break existing models.


joao-silveira (approved):
Looks good 👍

PR #2172, opened 2021-05-06
doc.markdown: First markdown language draft

joao-silveira:
This commit has two parts:

* A markdown language implementation including TextGen.

This implementation covers the basic syntax as described in:
https://www.markdownguide.org/cheat-sheet/ and additionally allows
including xml elements which allows mixing html elements with markdown
syntax.

* A generator from documentation language to markdown

In the long term we plan to replace the existing html generator with the
markdown generator. For that reason the rendered markdown is on purpose
very close to the existent html generator output.

For the cases where a documentation language feature is not covered by
the markdown basic syntax we just use html.


coolya (approved):
lets go! 🚀

PR #2174, opened 2021-05-11
Replace deprecated child of ListOfModelsScope

AlexeiQ:
Replace deprecated child of ListOfModelsScope:
The child with deprecated concept `models: ModelReferenceExpression` results in warnings in assessments, when ListOfModelsScope is used. To get rid of the warnings, the models-child with the deprecated concept `ModelReferenceExpression` is replaced with a current implementation: `ModelIdentity`.
Additionally a migration is added to migrate ListOfModelsScope.models children instances.

### DIFF
--- a/code/languages/com.mbeddr.analyses/_spreferences/CodeReviewPreferences/module.msd
+++ b/code/languages/com.mbeddr.analyses/_spreferences/CodeReviewPreferences/module.msd
@@ -12,7 +12,7 @@
   </facets>
   <sourcePath />
   <languageVersions>
-    <language slang="l:d4280a54-f6df-4383-aa41-d1b2bffa7eb1:com.mbeddr.core.base" version="5" />
+    <language slang="l:d4280a54-f6df-4383-aa41-d1b2bffa7eb1:com.mbeddr.core.base" version="6" />
     <language slang="l:7d323e61-8358-4656-a071-a2bb68438615:com.mbeddr.core.codereview" version="0" />
     <language slang="l:63e0e566-5131-447e-90e3-12ea330e1a00:com.mbeddr.mpsutil.blutil" version="1" />
     <language slang="l:d3a0fd26-445a-466c-900e-10444ddfed52:com.mbeddr.mpsutil.filepicker" version="0" />
--- a/code/languages/com.mbeddr.analyses/_spreferences/PlatformTemplates/module.msd
+++ b/code/languages/com.mbeddr.analyses/_spreferences/PlatformTemplates/module.msd
@@ -12,7 +12,7 @@
   </facets>
   <sourcePath />
   <languageVersions>
-    <language slang="l:d4280a54-f6df-4383-aa41-d1b2bffa7eb1:com.mbeddr.core.base" version="5" />
+    <language slang="l:d4280a54-f6df-4383-aa41-d1b2bffa7eb1:com.mbeddr.core.base" version="6" />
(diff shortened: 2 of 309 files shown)


AlexeiQ:
@wsafonov As we've discussed, I've tried to run the migration for this change as well on the project `com.mbeddr.mpsutil`. But the migration-wizard didn't offer me to execute the migration. 
I've rebuild the project and tried the migration-wizard, after:
1) opening the mpsutil project alone (without any other project)
2) opening `com.mbeddr.core` first and additionally the `com.mbeddr.mpsutils` project.
With the core project the wizard immediately spawn and wanted to migrate, while mpsutils wizard didn't want to execute anything.

sergej-koscejev (commented):
I'd suggest to use ModelIdentity rather than ModelPointerExpression. ModelPointerExpression is a base language Expression, whereas we just need a model pointer and should be conceptually independent from base language.

However, when using ModelIdentity it's necessary to supply a custom cell menu to show all available menus. I think the best way to do this is to create a menu contribution like AllRepoModels_Contrib ([MPS link](http://127.0.0.1:63320/node?ref=r%3A62b685a8-caa0-436e-bd8f-043142e659b2%28jetbrains.mps.build.mps.testManifest.editor%29%2F3815448882692900440)).

sergej-koscejev:
> With ModelPointerExpression at least we are not getting any worse and not adding any new dependency.

Since you're touching that part you could as well improve it to be state of the art and not stop halfway. Improving it later would mean another migration.

> I would check with JB why they don't use that menu on ModelPointer which would become then ideal fit for baselLanguage independent scenarios.

The reason could be that the menu shows all models in the repository whereas you might want to filter something out. And perhaps it's problematic to remove unwanted entries from the completion menu. But yes, it might be worth checking that.

AlexeiQ:
Thank you both for the valuable input. I'm going for the proposed state-of-the-art implementation now.

AlexeiQ:
* I've changed the implementation to the suggestion of @sergej-koscejev, adapted the migration and run the migration via migration wizard in @wsafonov s BigProject.
* The wizard didn't start automatically, because `c.m.mpsutil.actionsfilter` couldn't be deployed. 2021-06-01_12h50_18
* I triggered the migration wizard manually. 
* It stopped for two errors in solutions `com.mbeddr.rcp` and `com.mbeddr.rcp.actions`, which I ignored.
* Following migrations were executed: 2021-06-01_14h06_03
* I checked the changes done by the migration, nothing out of the ordinary

wsafonov (changes requested):
In general looks good now, just minor things:
- typo in behavior: `modelIdentity` instead of `modelIdenty`
- for the sake of uniqueness, let's give the contribution a unique suffix containing the concept name: `AllReposModels_MenuContrib_ListOfModelsScope`

Tip: you can use interactive rebase in MPS to quickly edit the required commit during rebase w/o the need to touch/change the commit with the applied migration.

AlexeiQ:
Integrated @wsafonov s feedback. Thx a lot for it :)

wsafonov (approved):

sergej-koscejev:
Should this be merged before or after the 2021.1 migration?

AlexeiQ:
This is for Lodas, thus I would merge it now = before the 2021.1 merge, in case Lodas doesn't want to migrate to 2021 asap.
@wsafonov / @sergej-koscejev  Is that okay with you?

AlexeiQ:
Got an okay from @wsafonov to merge 👍

PR #2177, opened 2021-05-27
Migrate to MPS 2021.1

wsafonov:
This PR includes all required changes for MPS 2021.1 version and migrations executed for all projects:

image

Prior to migration all build scripts have been reloaded to make sure that model roots for all modules are reflected correctly in the build models.


sergej-koscejev:
In commit 1cb62a104461298fea348a5d76ceee730bb09a36 @artem-tikhomirov has removed read action from the generator of parameterizedMenu while removing usages of `ModelAccess.instance()`. I doubt that's correct. The repository can be obtained via `editorContext` (possibly via `_context.getEditorContext()`).

sergej-koscejev (changes requested):
See my comment re read action in parameterizedMenu generator

artem-tikhomirov:
JFTR, I tried to get help from the author of parameterizedMenu code to discuss the change but didn't get any reply.
I don't see a reason to make explicit model read on per-action level when a menu is built. If there's a code that is invoked by MPS while building the menu and MPS doesn't get proper model read, we'd better fix this in MPS, rather than force each individual action to deal with model access on its own.

artem-tikhomirov:
It looks like there's a read action for each parametrized parameter. To me, the place that starts parameter evaluation shall grab proper access. I don't know if it's MPS code or some runtime of the parameterizedMenu language. I suspect it's getParameters(TransformationMenuContext) method in template case_TransformationMenuPart_ParameterizedSubMenu_declare, but don't know for sure.

sergej-koscejev:
@abstraktor I think you are the abovementioned "author of parameterizedMenu code", can you comment?

wsafonov:
After looking at the `parametrizedMenu` language I'm asking myself why cannot we support these scenarios (and maybe even more) in the MPS editor language? AFAICS, `parametrizedMenu` simply adds the possibility to put `submenu` into the `parameterized` transformation menu. This is obviously done as a workaround because the MPS editor language is currently too restrictive and only allows `actions` as parametrized content. But I don't see any reasons to not allow `submenu`, `parametrized` itself and maybe other menu elements like `group`, `include` inside `parametrized` menu (looks like they just need to implement `IParameterizableMenuPart` concept interface, maybe some minor generator adjustments would be needed as well). So maybe instead of figuring out, where a read action should be added to enable a workaround language, we should better support the actual scenarios from this language in MPS and deprecate the language in favor of the standard MPS language?

artem-tikhomirov:
Agree, if there's a way to accomplish the task with regular MPS facilities, we'd rather have this fixed in MPS.
I'm eager to advocate for the change in MPS, but not familiar with editor/actions enough to craft the change myself.

sergej-koscejev:
A challenging scenario would be supporting multiple parameter objects in scope, because there is only one `parameterObject` concept. So parameterized parameterized would need some thought. Parameterized submenus are fine, as the mbeddr implementation shows, and they definitely could be part of MPS. But what does this mean for this PR?

wsafonov:
> But what does this mean for this PR?

So if we all agree to move this language extensions to MPS (including solving possible nested `paramaterObject` issue), to proceed with this PR we could simply revert those changes and hope that MPS can ship those extensions in one of the next 2021.1.x bugfixes? I can create an MPS issue for that. As soon as the new version is released, we would deprecate the `parametrizedMenu` language.

sergej-koscejev:
> to proceed with this PR we could simply revert those changes

This would mean we keep using deprecated code, and I don't want to waste Artem's cleanup efforts.

wsafonov:
> This would mean we keep using deprecated code,

Yes, but this would only affect that single change in the language that is anyway not broadly used so far, until we can switch (migrate?) to the MPS editor language, which hopefully would happen rather sooner than later. I don't won't to invest more time now into something that we would anyway soon remove. @artem-tikhomirov how do you see this?

artem-tikhomirov:
Frankly, I don't like the idea of bringing MA.instance() call back. I'm trying to get rid of this singleton for years. I don't quite understand who and when gonna address the change, and I can easily imagine us discussing this call again in a couple of years. Is there's a sample use of the language I can check to see if there's a better place to put model read to? I'm pretty sure there's a way to have model read obtained in a place where there's editorcontext/repository instance available. I don't insist we have to avoid model read at all costs, I'd be satisfied if we just don't use MA.instance(). E.g. using editorContext.repository would be better.

artem-tikhomirov:
Unfortunately, the sandbox is quite rudimentary and doesn't help to understand the need for model read for QF_TM_ParameterizableParameter and QF_TM_ActionLabelText. There are no uses for these; all I can do is try to deduce possible scenarios from templates.
QF_TM_ActionLabelText provides body of ActionItemBase.getLabelText(String) in MPS sources. There's no explicit contract, however, at least few of the calls of the method provide read access. The method seems to be part of code completion process that runs with model read. Therefore, I see no reason to add explicit model read for this case. If there's any model read access error while accessing action item text, that would be a general MPS defect.

QF_TM_ParameterizableParameter provides body in 2 cases. First, case_TransformationMenuPart_ParameterizedSubMenu_declare, getParameters() method (in c.m.mpsutil.parameterizedMenu generator). The method doesn't override anything from MPS-supplied runtime class and its  only use is in commented out section of getParts() method in the same template. 
Second, the query may get processed in TransformationMenuPart_Parameterized_declare from j.m.lang.editor generator, where it overrides ParameterizedMenuPart.getParameters() method. I see a lot of MPS overrides of the same method that don't grab model access but access models/nodes, I'm pretty confident MPS builds a menu with proper model access, and I see no scenario these particular queries may get executed not inside model read.

sergej-koscejev:
I would introduce the contract that `getLabelText()` has to be invoked with read access, because the label text often needs to access nodes. I think the same applies to any method that returns item information or creates items. I checked the context actions tool and the context assistant in MPS and they seem to query items inside a read action. 

In that case we can leave the PR as it is.

sergej-koscejev (approved):

abstraktor:
Sorry folks, especially @artem-tikhomirov for being unresponsive.

As far as I can see, you're right with your assumptions about the idea of the language. I only know of YSEC as being a user of that language, but share the idea that this would ideally be fixed within MPS editor language, and then the parameterizedMenu language could die in peace.

For now, I can live with both, either leaving the model access in or out. We'll take care of adding the read when using the language if necessary.

artem-tikhomirov:
> I would introduce the contract that getLabelText() has to be invoked with read access, because the label text often needs to access nodes.
Totally agree. Now it just seems to be an implicit assumption, at least for major scenarios (like code completion menu). If there are some cases when MPS invokes supplied code without proper model access, that's MPS defect. IMO, language designer shall not care about model access at all when they provide contributions to MPS code.

wsafonov:
@artem-tikhomirov I have created a new issue for the missing transformation menu features in MPS editor language: https://youtrack.jetbrains.com/issue/MPS-33464

PR #2182, opened 2021-08-03
Show NodeAttributes in "embed model as image" paragraphs in mbeddr.doc

kbirken:
Currently NodeAttributes are not shown in model nodes which are added to mbeddr.doc documentation using the "embed model content as image" mechanism. This is due to the querylist flag `duplicates safe` being set to true in the two editors of the corresponding concept.

This PR sets the flag `duplicates safe` to false, leading to NodeAttributes becoming visible again.

Remark: Due to @slisson, problems may occur if the same node is being projected by querylist twice in the same editor and edited. In the current implementation, this may happen if the user creates two "embed model content as image" paragraphs for the same model node (or for a node and additionally for one of its descendants). However, I think this is unlikely in documentation. 

@joao-silveira We could also set the querylist in "embed model content as image" to read-only to avoid this. What do you think?

### DIFF
--- a/code/languages/com.mbeddr.doc/languages/com.mbeddr.doc/languageModels/editor.mps
+++ b/code/languages/com.mbeddr.doc/languages/com.mbeddr.doc/languageModels/editor.mps
@@ -8341,7 +8341,6 @@
           </node>
         </node>
         <node concept="s8t4o" id="627_yy34Gwz" role="3EZMnx">
-          <property role="28Zw97" value="true" />
           <ref role="28F8cf" to="tpck:gw2VY9q" resolve="BaseConcept" />
           <node concept="s8sZD" id="627_yy34Gw$" role="sbcd9">
             <node concept="3clFbS" id="627_yy34Gw_" role="2VODD2">
@@ -8748,7 +8747,6 @@
               <property role="VOm3f" value="true" />
             </node>
             <node concept="s8t4o" id="627_yy34GyM" role="3EZMnx">
-              <property role="28Zw97" value="true" />
               <ref role="28F8cf" to="tpck:gw2VY9q" resolve="BaseConcept" />
               <node concept="s8sZD" id="627_yy34GyN" role="sbcd9">
                 <node concept="3clFbS" id="627_yy34GyO" role="2VODD2">
--- a/code/languages/com.mbeddr.doc/solutions/com.mbeddr.doc.test.documents/models/code.mps
+++ b/code/languages/com.mbeddr.doc/solutions/com.mbeddr.doc.test.documents/models/code.mps
@@ -2,6 +2,7 @@
 <model ref="r:5de29430-c053-484a-9826-87e7556eee3e(com.mbeddr.doc.test.documents.code)">
   <persistence version="9" />
   <languages>
+    <use id="f2801650-65d5-424e-bb1b-463a8781b786" name="jetbrains.mps.baseLanguage.javadoc" version="2" />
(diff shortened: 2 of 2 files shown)


kbirken:
This PR is targeted at maintenance/mps20203, as we need it for a customer project. I already prepared a branch for 20211 (current master) and will create a PR for this as soon as this PR here has been merged.

coolya:
> Remark: Due to @slisson, problems may occur if the same node is being projected by querylist twice in the same editor and edited. In the current implementation, this may happen if the user creates two "embed model content as image" paragraphs for the same model node (or for a node and additionally for one of its descendants). However, I think this is unlikely in documentation.

Unfortunately this use case is quite common when it comes to documentation. Often people will use one node and then embed it multiple times but with different filters in the include. That way they can have parts of the node shown and explained individually while only maintain one full fledged example.

slisson:
I now reproduced the issue. If you enable "duplicates safe" it's the same as if you would show the node as part of a reference cell. MPS decides to not show NodeAttributes is that case (see https://github.com/JetBrains/MPS/blob/4efcf0f517d24fe6c235609e79e72131d81e643d/editor/editor-runtime/source/jetbrains/mps/nodeEditor/updater/UpdateSessionImpl.java#L220).
If we disabling "duplicates safe" the editor engine would return the same EditorCell instance at all location and because a cell can only have one parent the cell will just disappear at all other locations.

kbirken:
> I now reproduced the issue. If you enable "duplicates safe" it's the same as if you would show the node as part of a reference cell. MPS decides to not show NodeAttributes is that case (see https://github.com/JetBrains/MPS/blob/4efcf0f517d24fe6c235609e79e72131d81e643d/editor/editor-runtime/source/jetbrains/mps/nodeEditor/updater/UpdateSessionImpl.java#L220-L223).
> If we disabling "duplicates safe" the editor engine would return the same EditorCell instance at all location and because a cell can only have one parent the cell will just disappear at all other locations.

Would it be safe to set `duplicates safe` to false and set the querylist cell of "embed model content as image" to read-only?

slisson:
No, because then the node can only be rendered at one location. If there are two locations then at one of them the node will not be shown at all.

kbirken:
> No, because then the node can only be rendered at one location. If there are two locations then at one of them the node will not be shown at all.

I just did some tests with `duplicateSafe=false` and the queried model node of `ModelContentAsImageParagraph` being editable, exactly the configuration which is on the branch (tested with 2020.3.4). I couldn't reproduce the case where a node being referenced by more than one `ModelContentAsImageParagraph` nodes in the same root was not shown. So in theory it shouldn't work, in practice it looks good. So should we merge this PR anyway? Or are there obvious cases where nodes referenced twice will not be shown in one of the contexts?

In general, this problem with `ModelContentAsImageParagraph` (and similar usages of querylist) will be triggered by the end-user of the language, and the querylist implementer can do nothing about it. The only way would be to forbid referencing nodes which conflict with nodes already shown by querylists in the same root. This seems to be like a problem we should solve independently of the mbeddr.doc language... 

Comments?

slisson:
> I just did some tests with `duplicateSafe=false` and the queried model node of `ModelContentAsImageParagraph` being editable, exactly the configuration which is on the branch (tested with 2020.3.4). I couldn't reproduce the case where a node being referenced by more than one `ModelContentAsImageParagraph` nodes in the same root was not shown. 

Then we can merge the change.

joao-silveira (approved):

PR #2183, opened 2021-08-09
Show NodeAttributes in "embed model as image" paragraphs in mbeddr.doc

kbirken:
See merged PR #2182 for a discussion on this change. This is the change for current master, i.e., 2021.1. The PR #2182 was the same change for 2020.3 maintenance.


kbirken:
Merging, as this is the same change as in PR #2182, only on different branch.

PR #2184, opened 2021-08-12
doc.gen: fix in images export

joao-silveira:
Copy images for configs located in a different solution than the export configuration.


kbirken (approved):
Small change, big effect: My complex customer use case is working now (where the docconfig is in a separate solution than the document being rendered).

kbirken:
Pls port this fix also to master.

PR #2185, opened 2021-09-01
Fix ClassNotfoundException in preference components

alexanderpann:
When opening preferences, the following error message appears:
`java.lang.ClassNotFoundException: com/mbeddr/mpsutil/breadcrumb/runtime/plugin/BreadcrumbPreferences_Configurable`
This commit removes the offending line in AbstractPreferenceFormConfigurable.

The error happens because in Intellij the class [EpBasedConfigurableGroup](https://github.com/JetBrains/intellij-community/blob/0e2aa4030ee763c9b0c828f0b5119f4cdcc66f35/platform/platform-impl/src/com/intellij/openapi/options/ex/EpBasedConfigurableGroup.kt#L133) tries to load the class AbstractPreferenceFormConfigurable but fails. If anyone knows more about this, you can suggest a better
solution than commenting out the line that doesn't work.


coolya:
@alexanderpann its fine breadcrumbs is anyway not used at all and all the runtime is disabled for a long time because of performance. 

If you like you can sed an additional pr removing it entirely.

alexanderpann:
I know. After submitting the pull request, I noticed in another issue that they are deprecated but this exception happens every time I open the preferences which is quite annoying. That's why I wanted to fix it in the first place.

PR #2180, opened 2021-06-23
build: additional publishing to github packages

coolya:
The build now publishes the master and maintenance/* branches to github
packages in addition to the normal itemis maven repository. The changes
are backwards compatible and don't require any changes on teamcity,
apart from providing the github logins data.

Read the inline comments carefully because some changes are none obvious
without. Mostly caused by the fact that we configure our subproject from
a parent project was tricky to get right.

PR #2186, opened 2021-09-09
build: additional publishing to github packages (maintenance/MPS20203)

nkoester:
Applies identical changes as in PR #2180 to the MPS20203 maintenance branch so that this older version is also published to github packages.

See also: 2c585c7c815884659d360a548df9edea8dd5db67

### DIFF
--- a/build/com.mbeddr/build.gradle
+++ b/build/com.mbeddr/build.gradle
@@ -62,7 +62,7 @@ String minorVersion = '-Dminor.version=' + ext.mbeddrMinor
 ext.mbeddr_home = ['-Dmbeddr.github.core.home=' + file(rootProject.projectDir.absolutePath).getAbsolutePath(), minorVersion, majorVersion, buildVersion]
 ext.slall_home = ['-Dsl-all.home=' + file(artifactsDir.absolutePath + '/de.itemis.mps.extensions').getAbsolutePath(), minorVersion, majorVersion, buildVersion]
 ext.dependsOnMPS_scriptArgs = [mps_home, build_dir, artifacts_root]
-ext["itemis.mps.gradle.ant.defaultScriptArgs"] = ext.dependsOnMbeddr_scriptArgs = [*dependsOnMPS_scriptArgs, *mbeddr_home, *slall_home]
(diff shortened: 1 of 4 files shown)


build/com.mbeddr/build.gradle
```diff
@@ -62,7 +62,7 @@ String minorVersion = '-Dminor.version=' + ext.mbeddrMinor
 ext.mbeddr_home = ['-Dmbeddr.github.core.home=' + file(rootProject.projectDir.absolutePath).getAbsolutePath(), minorVersion, majorVersion, buildVersion]
 ext.slall_home = ['-Dsl-all.home=' + file(artifactsDir.absolutePath + '/de.itemis.mps.extensions').getAbsolutePath(), minorVersion, majorVersion, buildVersion]
 ext.dependsOnMPS_scriptArgs = [mps_home, build_dir, artifacts_root]
-ext["itemis.mps.gradle.ant.defaultScriptArgs"] = ext.dependsOnMbeddr_scriptArgs = [*dependsOnMPS_scriptArgs, *mbeddr_home, *slall_home]
+ext["itemis.mps.gradle.ant.defaultScriptArgs"] = ext.dependsOnMbeddr_scriptArgs = [*dependsOnMPS_scriptArgs, *mbeddr_home, *slall_home, "-Dmps.generator.skipUnmodifiedModels=true"]
```
coolya, line 65: do not add `-Dmps.generator.skipUnmodifiedModels=true` it will break the build. 

There is a long standing MPS issues that prevents us from using incremental generation in mbeddr. Unfortunately it's still not fixed. 

The build will fail with no obvious error message is there are no changes and the models.

coolya (changes requested):

build/com.mbeddr/build.gradle
coolya: At least in 2020.3 this is still a problem. On master with 2021.1 it seems to work.

nkoester: Sure!

coolya (approved):

Issue #2191, opened 2021-09-14
mbeddr.doc: Incorrect computation of stable-ids


kbirken:
Stable ids are used by mbeddr.doc to create links and references in resulting documents. However, the current implementation of stable ids has some issues:

- stable ids should be computed after resolving DocumentIncludes, not before
- if (maybe invalid) stable ids exist from previous generator runs, they will not be recomputed - the old ones will be used
- if getStableId is being called without a previous setStableId, the stable id is computed, but not cached

kbirken:
Resolved via PR #2192.

PR #2192, opened 2021-09-14
mbeddr.doc: Fix ToC link clashes and image file name clashes

### LINKED ISSUE #2190: mbeddr.doc: Non-unique names for ModelContentAsImageParagraph nodes leads to overwritten files
If two or more ModelContentAsImageParagraph nodes in a document have the same name property, then the generated images will have the same name, too. This collision leads to overwritten files and thus to wrong images in the resulting documents.

This case will most likely happen in conjunction with DocumentIncludes.

kbirken:
Due to problems with stable-id computations (see issue #2191), duplicate stable-ids have been generated for different link targets. This is solved by this PR. Additionally, image file names could be not unique if two or more `ModelContentAsImageParagraph` nodes have the same name (issue #2190). This is also fixed by this PR.

This PR closes #2190 and #2191.


joao-silveira:
@kbirken i've included the sections from included documents in the TOC presentation mode, can you have a look?

joao-silveira (approved):

PR #2197, opened 2021-09-22
Adds a workaround to prevent StackOverflow (issue #2196)

dbinkele:


joao-silveira:
@alex-y-wd Hi! It seems like in some conditions `setSelectedContent()` triggers an ancestor added event which leads to a stackoverflow. Given that you (as far as i can see in git history) implementing the solution could you give us some feedback? does this looks live a good solution? do you have a better idea?

joao-silveira:
@dbinkele from what I can see the event is triggered when there is a change in the selection. This happens when there is both content for the concept and node editor components, first the node editor component is selected and afterwards the concept editor component. When the selection changes and the concept editor component is selected the AncestorEvent is triggered. 

To be honest I don't know the difference between `setSelectionContent` and `addContent` so i can't tell if this is the right fix or not.

joao-silveira (approved):

PR #2198, opened 2021-09-22
Maintenance/mps20193 seca

abstraktor:
as discussed with kolja.

(to avoid version conflicts with `maintenance/mps20193…` branches)


alexanderpann:
@abstraktor I just saw that this PR was never merged. Except for the two commits for the actionsfilter language everything looks mergeable. I would only change the graphstream commit a bit so that it isn't a breaking change. Any objections to me merging those changes into 2021.1.4 since 2019.3 is deprecated?

AlexeiQ:
I'll close this PR and delete the branch, since the idea is to merge only the necessary changes via PR #2221 .

PR #2201, opened 2021-09-23
doc.gen: allow referencing documents from different models

joao-silveira:
It is possible to reference documents from different models in two
different ways:
* Using a document include checking the "reference only" box

This will contribute an entry to the table of contents that links to
the referenced document

* Using a doc ref word

This will generate a "word" link to the referenced document. Nothing is
added to the table of contents.


joao-silveira:
For an example see http://127.0.0.1:63320/node?ref=r%3Ac20abc99-c3e1-4a29-9270-de587f318ef3%28com.mbeddr.doc.test.documents.doc_refs%29%2F2671071626003091836

kbirken (changes requested):
Found two potential issues:

1. The "reference-only" document include is added to the ToC in the generated html, but not to the new ToC in presentation mode.
2. Often the top-level structure of a document is built of chapters instead of sections. If the "reference-only" document include contains a chapter, the generated ToC in the main document doesn't look nice. See screenshot:

image

("First" and "Second" are chapters as well).

kbirken (approved):
The two issues found last time are resolved now, LGTM.

PR #2204, opened 2021-10-18
platform artifact without mps-extensions

coolya:
build: add separate platform artifact

Added a new artifact to the mbeddr platform that doesn't repackage
mps-extensions. It exposes the dependency to mps-extensions via the pom.
This allows consumers of the artefact to decided which mps-extensions
version they would like to use.

This change breaks existing build scripts because of the new MPS build
solution that does only build the mbeddr platform. To fix this open the
build script in MPS and press F5 to fix all the references.

The artefact that repackaged mps-extensions is still present and remains
untouched. Down stream projects do not need to change their gradle scripts.

### DIFF
--- a/build/com.mbeddr/platform/build.gradle
+++ b/build/com.mbeddr/platform/build.gradle
@@ -126,7 +126,7 @@ task defaultWrapper(dependsOn: build_platform) {
 
 rootProject.defaultTasks 'defaultWrapper'
 
-task publishMbeddrPlatformToLocal(dependsOn: ['publishMbeddrAllScriptsPublicationToMavenLocal', 'publishMbeddrPlatformPublicationToMavenLocal'])
+task publishMbeddrPlatformToLocal(dependsOn: ['publishMbeddrAllScriptsPublicationToMavenLocal', 'publishMbeddrPlatformPublicationToMavenLocal', 'publishMbeddrPlatformSlimPublicationToMavenLocal'])
 
 publishing {
     publications {


sergej-koscejev:
Looks good but the breaking change is unfortunate. Since most projects are currently using 2020.3, can we copy the relevant parts in 2020.3 so that it remains compatible, and apply the breaking change to 2021.1 only?

coolya:
> Looks good but the breaking change is unfortunate. Since most projects are currently using 2020.3, can we copy the relevant parts in 2020.3 so that it remains compatible, and apply the breaking change to 2021.1 only?

I tried that first but got a bunch of "module visible from x and y" errors. I didn't investigate further why. I can take another look at this later. For me there is no urgent pressure right now to port this back to 2020.3. My initial problem with mps-extensions is fixed for now. Might be better to just to it in 2021.1 then and projects get change when they do the migration.

sergej-koscejev (approved):
Looks good, I would perhaps add `publishMbeddrPlatformSlimPublicationToMavenLocal` to `publishMbeddrPlatformToLocal` task on line 129.

alexanderpann:
@sergej-koscejev do we still want this PR to be merged?

sergej-koscejev:
> @sergej-koscejev do we still want this PR to be merged?

Yes. I think we may want to revisit the names later (I would make the "slim" artifact the default and have a qualifier for the "fat" one: -bundled, -all, or something like that) but I think we can merge it to move forward.

sergej-koscejev:
We still want this but will have to reimplement it from scratch anyway because things have changed in four years. I have created #2834 to track that and will close this PR.

Issue #2211, opened 2022-03-04
Refactor Method 'getAllImports()' in Chunk.


dbinkele:
In order to fix a wrong behavior in iets3.core (https://github.com/IETS3/iets3.core/pull/860) we need better control of this method. Especially, if during the run where dependencies are collected transitively, only those with reexport flags should be considered, or all. 
This is enhanced control should be achieved by this ticket.

dbinkele:
done #2212

PR #2212, opened 2022-03-04
Enhance control of 'allImports'-method (#2211)

dbinkele:
#2211


kbirken (approved):
LGTM. The original handling of reexported imports and the new handling are somehow tricky, but after some code inspection I believe the original behavior hasn't been changed by this PR (which is good).

PR #2220, opened 2022-06-02
Document aspect: add priorities 

alexanderpann:
This PR adds support for setting priorities for the annotations of the  documentation aspect. Now it is possible to document concepts of other languages in the documentation aspect as well. The annotation with the highest priority is used when there are multiple available documentations for a concept.

### DIFF
--- a/code/languages/com.mbeddr.doc.aspect/languages/com.mbeddr.doc.aspect.exampleLanguage.extended/models/documentation.mps
+++ b/code/languages/com.mbeddr.doc.aspect/languages/com.mbeddr.doc.aspect.exampleLanguage.extended/models/documentation.mps
@@ -17,6 +17,7 @@
       </concept>
       <concept id="7810506636291686467" name="com.mbeddr.doc.aspect.structure.DocumentedPropertyItemAnnotation" flags="ng" index="fANS$" />
       <concept id="1058510331725720478" name="com.mbeddr.doc.aspect.structure.DocumentedConceptAnnotation" flags="ng" index="3n9NSn">
+        <property id="1881564090922902400" name="priority" index="17ySGi" />
         <reference id="1058510331725761196" name="concept" index="3nadW_" />
       </concept>
     </language>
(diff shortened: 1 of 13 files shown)


AlexeiQ:
The priorities work fine on sections (Concept/Property - Annotation) and list-items (PropertyItemAnnotation) and can be also used to show a concept-documentation instead of a property/PropertyItem-documentation, if the priority is set higher 👍 .

I suggest 2 code-simplifications:
1. The new `foreach-loop` in [`DocumentationAspectHelper.findDocumentationElementSingle()`](http://127.0.0.1:63320/node?ref=r%3Ae117f55c-1f24-4b31-a4cc-7557b8737f3e%28com.mbeddr.doc.aspect.runtime%29%2F1881564090925389780) is obsolete, because the last `foreach-loop` already covers this withe [it's first `if`](http://127.0.0.1:63320/node?ref=r%3Ae117f55c-1f24-4b31-a4cc-7557b8737f3e%28com.mbeddr.doc.aspect.runtime%29%2F86363842538987197).
2. In the same method [`set<node<Document>> documents = collectDocuments(model);`](http://127.0.0.1:63320/node?ref=r%3Ae117f55c-1f24-4b31-a4cc-7557b8737f3e%28com.mbeddr.doc.aspect.runtime%29%2F1881564090925382220) can be moved out of the `if` and thus only be calculated once for all `foreach-loops`.

AlexeiQ:
Added code-simplifications as proposed before:
1. Instead of removing the newly added `foreach`-loop as suggested, I removed the `if(property != null)...`-part in the last loop. Thus, if we are looking for a `property`-documentation, we first look for the documentation on all `Items` and then on all `Sections`. If no documentation is found, we look for a `concept`-documentation. If the new loop would be removed instead and we would be looking for a property and would not have found an documentation on `Item`, we would find either the FIRST `property`-documentation OR `concept`-documentation, because we would have looked for both in each `Section`. Since usually the concept is documented before it's properties, we would usually find the concept-documentation, even if we would be looking for a property-documentation. Now, in case we are looking for a property-documentation, we look at first for a property-documentation and take the concept-documentation only if we don't find a property-documentation.
2. Moved the var as proposed.

alexanderpann:
@arimer You are right. I didn't think that anyone would use this method outside the documentation aspect, that's why I didn't consider deprecating the old one.

Noted 👍

PR #2225, opened 2022-07-21
mbeddr.core.build: add version number to mpsutil.uisupport

arimer:
the same versioning number is now used as for all the other plugins


arimer:
Thx! @slisson  👍🏼

PR #2240, opened 2022-10-31
Documentation: generate images in own facet

### LINKED ISSUE #2239: Facet for generating screenshots
Screenshots are currently being generated by com.mbeddr.doc.generator.template.main.renderCodeScreenshotsAndAttachOutputLocation ([MPS link](http://127.0.0.1:63320/node?ref=r%3A716d4aa1-57d8-4a9c-b02d-c1375c64785d%28com.mbeddr.doc.generator.template.main%40generator%29%2F8730648445433126395)) but it does not work properly from the command line. The generator runs under a model lock on a background thread, and screenshots have to be taken in EDT and need their own lock. This is prone to deadlock.

A better solution would be to take screenshots from a make facet because they run outside of any model locks. There is an example of such facet in MPS in language jetbrains.mps.lang.editor.imageGen that we could use.

alexanderpann:
- fixes #2239
- fail the generation when an image can't be generated instead of silently failing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,22 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). The project does _not_ follow
+Semantic Versioning and the changes are simply documented in reverse chronological order, grouped by calendar month.
+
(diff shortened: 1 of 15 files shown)


sergej-koscejev:
I have changed the user object approach to use TracingUtil from the MPS generator and fixed the term dependencies visualization.

PR #2242, opened 2022-11-02
Migrate to MPS 2021.3.1

alexanderpann:


wsafonov (changes requested):
In general looks good for me. Just minor stuff:
- what's the expectation for the list of versions in readme? It looks like it claims that those versions would be officially supported, which won't be the case. Either we should describe the list differently or reduce it just to the list of officially maintained versions
- changes in the BigProject project metadata should be reverted to keep those files clean

As soon as MPS 2021.3.2 version is available, we can add it to this PR and merge.

alexanderpann:
- At the time of writing the readme, the list reflected the versions that were used (yes, 2019.3 was still in use). Yes, I think we should only list the versions that are supported. What about supporting the same versions as in IETS3? (Master + last 3 maintenance).
-  Sure.

PR #2250, opened 2023-02-06
xml.fix: work around incorrect XML escaping in MPS 2021.3.2

sergej-koscejev:
Tracked under MPS-35321/MPSSPRT-409 in MPS. The bug is fixed in 2022.2 so the workaround will not be needed there.


alexanderpann (approved):
LGTM. To fix the warning in `test.com.mbeddr.mpsutil.xml.fix.support` use the design scope for the dependency on the sax generator. [It is the recommended way](https://www.jetbrains.com/help/mps/the-generator-algorithm.html#dependencyscope/kind-'generationtarget'and'design'.) when referencing another generator to specify priority rules.

alexanderpann (approved):

sergej-koscejev:
The checks are failing because of some itemis infrastructure issues. I'm merging the PR anyway.

PR #2255, opened 2023-02-27
mpsutil.json: added paste and import action

arimer:
This PR picked up already existing work from the branch _feature/json_importer_ and finalizes it. The behaviour of the paste and import actions is similar to the one provided by for the `xml` laguage.

#### JSON paste action

Inside an **empty JSON document** or inside an **JSON object**,  users can paste valid and parsable JSON object text-snippets.

grafik

#### JSON import action
In addtition to the paste we also allow imports of **multiple JSON files**. Therefore a model action is provided allowing the users to select multiple json files for import.
grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 * Trace explorer: don't try to highlight a node when there is no node in the tool.
 * Trace explorer: Fix exception that occured when the rerun button is pressed when there is no trace.
 * Fix exception `NullPointer: Cannot create configurable` in userstyles language.
+* json: Added paste and import action for json snippets and files.
 
 ## com.mbeddr.core.base
 
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -20211,8 +20211,8 @@
             <ref role="3bR37D" to="ffeo:1TaHNgiIbJb" resolve="MPS.Platform" />
           </node>
         </node>
-        <node concept="1SiIV0" id="10EaF0SRU33" role="3bR37C">
-          <node concept="3bR9La" id="10EaF0SRU34" role="1SiIV1">
+        <node concept="1SiIV0" id="mK9TSm5HmT" role="3bR37C">
+          <node concept="3bR9La" id="mK9TSm5HmU" role="1SiIV1">
             <ref role="3bR37D" to="ffeo:1TaHNgiIbIQ" resolve="MPS.Core" />
           </node>
         </node>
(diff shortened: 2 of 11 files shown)


alexanderpann (changes requested):
I've selected a folder in the file chooser and pressed ok:
```
java.lang.NullPointerException
	at com.mbeddr.mpsutil.json.plugin.importJsonFile_Action$1.value(importJsonFile_Action.java:68)
	at com.mbeddr.mpsutil.json.plugin.importJsonFile_Action$1.value(importJso
```
You don't stop the parsing, when an invalid file was selected (e.g. test.js) -> fatal error. I have also selected a wrong file, and got an empty error message.
To avoid hard-coded strings, you could use `jetbrains.mps.tool.common.PathManager.getUserDir`.
Reimporting the same file after making a change doesn't work, the first version is imported again.
Not sure if this is on purpose, but the action is always active, even if you don't use the JSON language in the model. The guard from the XML action is missing: `new ModelImports(this.mdl).getUsedLanguages().contains(language/jetbrains.mps.core.xml/)`
Again not sure if this is on purpose but escaped values such as \1234 are converted to unicode after the file is imported.

arimer:
>I've selected a folder in the file chooser and pressed ok:

I was not even aware that you could select folders. Looks like the configured FileDescriptor is not working as expected :(

>. I have also selected a wrong file, and got an empty error message.

What does wrong mean? A non .json file? If so then we can try to switch to an other file chooser, because my expectation would be that we should not even show anything else then .json files (and folders for navigation)


>Reimporting the same file after making a change doesn't work, the first version is imported again.

This seems weird. In case the content of the json file changed, I would expected the parser to pick up the new changes 🤔  Did you kept the "chooser window" open while updating the .json file?

>Not sure if this is on purpose, but the action is always active ... use the JSON language in the model.

Acutally, I is on purpose. I though that the action will only show up in case the langauge is used in the solution and was thinking that the actions are only applied after a dep. to the langauge is defined. Because the actions are part of a plugin, that is part of the `json.lang` 🤔  maybe I was wrong.

>.. escaped values such as \1234 are converted to unicode after the file is imported.

Tbh. I did not look in dept into the parser impl. but just picked up the work. Will have a closer look here.

alexanderpann:
I already mentioned it, but with the Mac native file chooser you can basically circumvent any filters, that's why you should check the choosing file/folder anyway.

arimer:
>Reimporting the same file after making a change doesn't work, the first version is imported again.

It seems like a bug in [FileChooser](https://github.com/JetBrains/intellij-community/blob/idea/223.8617.56/platform/platform-api/src/com/intellij/openapi/fileChooser/FileChooser.java). The inputstream / content of the selected file is not updated, although the file changed on file system. What a mess :/ 

I will switch to a Java swing based chooser instead.

sergej-koscejev:
> The inputstream / content of the selected file is not updated, although the file changed on file system. What a mess :/

The mess is probably a feature: VFS and file system caching.

arimer:
Yeah you are right I guess

> To provide the last two features, the VFS manages a persistent snapshot of some of the user's hard disk contents. The snapshot stores only those files which have been requested at least once through the VFS API, and is asynchronously updated to match the changes happening on the disk.

But at least it could update the snaptshot.

alexanderpann:
Can you please check why the build is failing? I also get a bad property index key failure in the plugin aspect.

arimer:
@alexanderpann 

- I've switched back to the vfs based chooser and added the required filterings and vfs refreshes. 
- A direct re-import of the changed files however is still not reflected directly after the import for me. I am not sure how else we can influence the refresh besides using the API for refresh
- a rebased also, so you might need to reset hard to the current head of origin.

alexanderpann (approved):

PR #2264, opened 2023-03-14
2253 Provide marker interfaces for logical child pattern

### LINKED ISSUE #2253: Provide marker interfaces for "logical child" pattern
# The "logical child" pattern

In several applications built with MPS we see the following pattern being used: Some contents logically belonging to some AST node are not stored as contained children, but are extracted to another root node instead. We will call the extracted content _logical child_, and the node it belongs to _logical child owner_.

The actual connection between child and owner is established as a reference from the logical child node to its logical owner (as suggested from the term "owner"). The owner does not have a direct reference to its logical children; in order to locate them, a module-wide scope is being used as a default. This allows to add various kinds of logical children to an owner without the owner needing to "know" them.

From the user interface point of view, this pattern is often called _aspects_ or _add-ons_.  It is usually implemented by providing tabs at the bottom of the editor window, using the same infrastructure as the language aspect tabs in MPS.

# Example application for this pattern

The artifact language might be a DSL for describing nested component structures (a component definition might contain sub-components, i.e., instances of other component definitions). The logical child extension then allows to add some specific information to each component definition (like parameters, capabilities, needed resources, etc). This additional information can be accessed by switching tabs at the bottom of the editor window (one "main" tab for the actual component definition, and one additional tab for each type of add-on information).

# Why provide marker interfaces for this pattern?

This pattern shows up in various independent applications, but it cannot be recognized explicitly in the code. Providing marker interfaces improves this, and brings the following benefits:

- Applications of this pattern could share common documentation/explanation, therefore it is more clear that this pattern is applied.
- In a second step, we could provide abstract helper concepts or even a standard implementation of this pattern using the usual "aspect tabs" approach.
- Based on the marker interfaces other generic languages and frameworks can provide special support for this pattern. The motivating example here is IETS3 Variability - see issue [IETS3.core 1017](https://github.com/IETS3/iets3.core/issues/1017) for more details.

# Implementation proposal

We propose two marker concept interfaces:

- `ILogicalChild` for concepts which are used with the logical child role (see pattern description above)
- `ILogicalChildOwner` for concepts which might be owners of logical children

Note: Any concepts implementing `ILogicalChild` or `ILogicalChildOwner` might be root nodes, but do not need to.

The minimal behavior aspect for these interfaces is a method which retrieves the owner for a given child:

```
concept interface ILogicalChild {
   node<ILogicalChildOwner> getOwner();
}
```

In a later step, we might add a method for the `ILogicalChildOwner` to retrieve all logical children. However, for the pattern this is not mandatory and might also be implemented by some auxiliary class. The retrieval involves searching a scope (e.g., a module or solution) for all matching logical children.

mfsoliveira:
This PR introduces 2 interfaces: 
- ILogicalChild: it marks a class that contains a reference to a node in another root.  
                        This reference can be a root itself. This reference is stored in the owner.  
                        Such reference creates a logical link between those nodes with the meaning logical child belongs to an owner. 
                        A logical child describes some aspect of the owner or adds information to it without forcing the persistence 
                        to be located in the same owner root. 
                       The method getOwner provides a default implementation to retrieve the owner of a child.

- ILogicalChildOwner: it delegates additional information to its logical children without forcing the persistence
                                   to be located in the same owner root. It does not know the existence of its logical children,
                                   thus a search method findLogicalChildren with a default implementation is provided.
                                   The default implementation searches in the solution of the owner inside existing models for 
                                   ILogicalChild that contains references to the owner.

This PR closes #2253 

## Checklist for creating a good PR (can be deleted)

- [x] I've described my change in the CHANGELOG.md if it is visible to users of the platform (**mandatory**).
- [x] I've selected the correct [base branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request).
- [x] I have checked my code (edge cases, no errors) and corrected any misspellings.
- [x] I have added comments to my code in areas where the code is hard to understand.
- [x] I've created a new language/solution.
    - [x] I've added the language/solution to the build script ([tutorial in readme](https://github.com/mbeddr/mbeddr.core#contributing-to-the-mbeddrplatform))
    - [x] I've added a sandbox that showcases the language.
    - [x] There is some kind of documentation available in the repository or somewhere else [e.g. [MPS Platform Docs](http://mbeddr.com/mps-platform-docs/platform_essentials/com_mbeddr_mpsutil/)].


kbirken (changes requested):
I fixed a dependency warning and updated the javadoc comments a little bit. Here are two findings which should be improved before merging:

- The `ILogicalClient` marker interface should not contain the "owner" reference as a structural member. Instead, the implementor of ILogicalClient should store this (using a more specific reference to its owning concept) and just provide it via the `getOwner` behavior method. 
- I would assume the implementation of `ILogicalChildOwner.findLogicalChildren()` should cope with modules in general, not restricted to _solutions_. For instance, a _language_ is a _module_, but not a _solution_. E.g., this would allow using the scheme in some utility-model inside a language definition. We sometimes use this pattern of utility-model when adding some documentation to a language, which also uses some example models of the language.

kbirken (approved):
Thanks for updating. Looks good now, approve from my side. 

I finally pushed one little commit to remove an unused private behavior method.

wsafonov (changes requested):
- it seems to me that the method for finding all models which could contain logical children (`getModelsFromOwnersModule()`) should be rather a strategy pattern with the default implementation allowing the specific owner to provide its own strategy. Let's make it virtual protected and rename to `getModelsWithLogicalChildren()`
-.we can safely remove empty generator modules as they won't be required
- I'm not quite sure about the selected name for the language. Maybe something more generic like logicalStructure would be a better choice? @kbirken what do you think?

wsafonov (approved):
Looks good, leaving the language names unchanged for now.

PR #2267, opened 2023-03-16
Modernize build to Gradle 8.0.2 and mps-gradle-plugin 1.13

sergej-koscejev:
Upgrade Gradle and mps-gradle-plugin so that the project can be built on an M1 Mac (Gradle 5.2 does not support M1 and the build fails because current directory is set to the daemon's working directory and not to the project directory).


sergej-koscejev:
We're hitting a Gradle bug here and I'll try reporting it and then working around it.

alexanderpann:
Do you mind explanation the bug?

sergej-koscejev:
I reported it at https://github.com/gradle/gradle/issues/24368

Gradle gets confused because several tasks output files into the same directory and Gradle doesn't like it. It's present in Gradle 7.6.1 as well.

sergej-koscejev:
This is now ready for review.

build.gradle
```diff
@@ -1,17 +1,10 @@
...
-        mavenLocal()
-        maven { url 'https://artifacts.itemis.cloud/repository/maven-mps/' }
-        mavenCentral()
-    }
-
-    dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.2.177.46341c0'
-    }
+plugins {
+    id 'de.itemis.mps.gradle.common' version '1.13.+'
```
alexanderpann, line 4: Where does this package name come from? How does this work?

alexanderpann (approved):
Is there a reason to only target the master branch?

build.gradle
sergej-koscejev: From here: https://github.com/mbeddr/mps-gradle-plugin/blob/v1.x/src/main/kotlin/de/itemis/mps/gradle/common.gradle.kts

It is an empty plugin that does nothing but causes Gradle to put its classes on the build script classpath.

sergej-koscejev:
> Is there a reason to only target the master branch?

I'm not getting paid to do this and I don't want to distract myself with all the branches and potential merge conflicts.

PR #2269, opened 2023-03-22
Canonicalize paths in the resource editor

HeikoBecker:
On Windows, the path uses by default a `\` instead of the `/` used by MacOS/Linux. The Windows platform Java can handle the forward slash at runtime. This commit ensures that only forward slashes are used in file paths.


alexanderpann:
Please don't forget to update the CHANGELOG.

PR #2272, opened 2023-03-31
Remove spawner and debugger

sergej-koscejev:
Remove unmaintained spawner and debugger projects from the mbeddr platform.

This is a breaking change.


eugenschindler:
Just a question: does this mean one cannot run and debug C code written in mbeddr anymore?

sergej-koscejev:
> Just a question: does this mean one cannot run and debug C code written in mbeddr anymore?

Yes. Were you using this functionality?

> What about starting other executables like CBMC or custom executables?

This still works as part of make, it doesn't use the spawner.

eugenschindler:
> Yes. Were you using this functionality?
Yes, but where we use that, we are on a much older MPS version, so maybe that's why I didn't notice that it's not maintained anymore.
It's not used on a critical path, but still used...

> > What about starting other executables like CBMC or custom executables?
Yes, but wasn't the spawner for non-build things? E.g. running plantuml? Or some assessment stuff?

sergej-koscejev:
> Yes, but wasn't the spawner for non-build things? E.g. running plantuml? Or some assessment stuff?

No, spawner was quite specific in that it could send signals to the processes and so on. Other programs are started using standard Java means.

wuryel:
I always considered running and especially debugging mbeddrC code directly from within MPS a key selling point. I used it in a demonstrator some time ago. Still I think engaging mbeddrC/MPS would be a feasible option to significantly improve our satellite flight software development at OHB and I hope to go this path in a near future project. Dropping the debugging feature does not sound very good to me.
BR / AndreasWortmann

eugenschindler:
That would be a good description of my case, too. It's not critical... yet, because I didn't yet get to spend proper attention to it so we use it more broadly.

alexanderpann:
@sergej-koscejev I know that the spawner is an annoying dependency for some subprojects to build, but why do you want to remove this functionality in the first place?

sergej-koscejev:
Isn't it being an unused and annoying dependency not reason enough? It made it difficult to reorganize build scripts, and I thought it was a failed experiment that could be removed by now.

alexanderpann (approved):
Alright. Since you already got the okay from @stephaneberle9 in Slack, I'll also approve.
I guess cleanup up this stuff make sense, since not enough people are using the debugger and spawner. Outdated jars, .so, .exe and .dll files are also not what you want to see in such a repository. It would be interesting to check if this has any effects on the needed libraries and executables in https://github.com/mbeddr/mbeddr.build.docker, but I guess not.

sergej-koscejev:
This has been closed for a while now but a few more reasons to remove spawner and debugger, now that I'm digging in the mbeddr source again:

1. Spawner and debugger are somehow special and make it impossible to simply open the entire mbeddr project (using tools/BigProject) and build it in MPS. Language com.mbeddr.core.debug cannot be loaded if spawner is not installed as a plugin.
2. Spawner contains some native libraries under `tools/mbeddr.spawner/lib/spawner` but it's not clear where they come from, there's no library for AArch64 Mac, and `macosx..x86_64` folder is likely misspelled (extra period), making me wonder how this thing could ever have worked on any Mac.

PR #2276, opened 2023-04-25
Bugfix/links between docs via doc ref word

mgronover:
A customer complains about broken builds when generating documents using mbeddr.doc.
The reason for was found in resolving remaining includes (see according mapping script), where all documents were removed that are not referenced by an include. 

grafik

Additionally the warning when generation the doc
grafik
was fixed.


alexanderpann (approved):
LGTM

PR #2278, opened 2023-05-18
ModelHelper#createModel(): prevent deadlock when called from EDT

sergej-koscejev:
Original code used an AsyncPromise and waited for its result but in some cases it would deadlock. I suspect the problem is that the code to set the promise would throw an exception and never set a result nor an error. This PR rewrites the code to use runInUIThreadAndWait which is guaranteed to terminate one way or the other.


sergej-koscejev:
My understanding of this is incorrect because runInUIThreadNoWait executes the Runnable directly when in EDT. I'm closing this for now and will reopen when I understand the problem correctly.

PR #2286, opened 2023-07-05
Remove some old/unused languages

alexanderpann:
The following languages will be removed (they have no usages in any of our platforms):

- `com.mbeddr.mpsutil.testScope` (superseded by [MPS test scopes](https://www.jetbrains.com/help/mps/testing-languages.html#testingscopes))
- `com.mbeddr.mpsutil.runconfiguration` (the workaround is included in MPS nowadays)
- `com.mbeddr.mpsutil.rcpconfig` (an internal module that is not needed anymore)
- `com.mbeddr.mpsutil.nodediff` (replacement: selected two nodes in the logical view, right-click -> Compare two nodes)
- `com.mbeddr.mpsutil.licensemanager.common` (an internal module that is not needed anymore)
- `com.mbeddr.mpsutil.langstats `(not very useful, was never meant as a productive language)
- `com.mbeddr.mpsutil.gradlesupport` (incomplete; use the language [com.dslfoundry.plaintextgen](https://jetbrains.github.io/MPS-extensions/extensions/generator/plaintext-gen) from MPS-extensions instead)
- `com.mbeddr.mpsutil.buildutil`(an internal module that is not needed anymore)
- `com.mbeddr.mpsutil.forms` (experimental language that was never completed)


svott7:
Hey @alexanderpann : I found the option to compare two nodes via editor popup menu useful and don't like the logical view option. There, members are usually not dispayed and sometimes I want to compare any node instead of root nodes only. What do you think about re-adding these actions? I could also copy the old features to my own MPS extensions.

alexanderpann:
You are right.  I guess we have to revert this change. The action works with non-root nodes too but the problem is that even if you add the action to the editor popup you can't select multiple nodes. I'll create a ticket. In the meantime you can add it to your own project or use the following snippet on the console: `StructDifferenceDialog.showNodeDifference(ProjectHelper.toIdeaProject(#project), leftNode, rightNode, "Left", "Right")`  
where `leftNode` and `rightNode` are copied references to nodes in the editor (for example: nodeRef@83515)

PR #2297, opened 2023-08-11
EditingGuide: Fix layout-bug in the task description

KemalSoysal:
Alexei Quapp's commit 672025e5db39445aa6cbd00fcddd773b7b4efd22 from branch refactor/mt-mps20203-with-seca-193 somehow was missed to be merged into the maintenance branch.

## Problem
Due to a layout-problem, the description-text was moved down below the hint and the buttons and thus overlapping with them.
Same goes for the hint text in some cases.

## Fix
Introduced `TopDownLayoutCells` to resolve the overlaying of cells.


KemalSoysal:
I reviewed the commit and cherry picked from the original branch in production.

PR #2305, opened 2023-09-07
Add tests for assessments

sergej-koscejev:
A few tests for assessments. I wanted to refactor them a bit but I got stuck so I thought I'd at least add the tests.


alexanderpann (approved):
Looks good. What do you want to refactor? I didn't check the implementation, the only code I know that's not good is the Project scope implementation.

sergej-koscejev:
We have some tests that update assessments and export the results. MPS 2022.2 complains when you update a read-only model (which is the case when running tests from Ant because the test solution is packaged into a .jar and executed from there).

There is a way to obtain query results without updating the model (using `AssessmentQuery#runQuery()`) but sometimes the previous results contain additional information such as a property to show they are acceptable, and for what reason, and I want to export that information too. So I wanted to implement a method to run the query, merge it with the current results, and return the merged result but without updating it.

But it may be easier to just copy the assessments to a temporary model and update them there.

PR #2306, opened 2023-09-08
Improve assessment editors

sergej-koscejev:
The goal was to improve the visual separation of individual assessments in an assessment container.

Before:

image

After:

image


sergej-koscejev:
I improved the alignment of the header a bit (it's 3 pixels too far to the right on the screenshot), foce-pushed.

alexanderpann (approved):
Good usage of color, and lines to separate headers from content and queries from each other.

Two further improvements that I could think of:

*   The folded cell of the assessment could show some more information e.g. the number of queries. The JetBrains way is to show  `{...}` but I also don't like that when we have enough horizontal space for information.
*   General problem: The swing buttons are made for ants and are way too small. The styling of them could definitely be improved.

Issue #2310, opened 2023-10-09
Update of "Context Action 2" should be performed even when not visible


mgronover:
The [Context Action 2 ](http://127.0.0.1:63320/node?ref=r%3A7db2f202-2a6e-4f79-9d01-79eb67397dc9%28com.mbeddr.mpsutil.contextactions.runtime.plugin%29%2F6294660000051114344) should be possible to be updated even when it is not shown or minimized.

Explanation:
If the context is changed when the ToolWindow is hidden no update is performed.
When the ToolWindow is getting visible again it shows the old content.

PR #2311, opened 2023-10-09
Added option to update when ToolWindow of context action even when invisible

mgronover:
**This addresses issue: #2310**
The context action 2 ToolComponent should be possible to be updated even if the window is not visible. This prevents the user from changing the focus to the components context again.

**The old behavior is used as the default one.**
If someone wants this new behavior a explicit call to a function has to be made.


alexanderpann:
How do you use the new method and how could I test the change? Why where the context variables changed (e.g. myNextContext != context)?

alexanderpann:
How about a node test case like this (untested):

```
tool<ContextActionsTool> tool = this.project.tool<ContextActionsTool>
ToolComponent cmp = (ToolComponent) tool.getComponent();
cmp.setRefreshOnlyWhenVisible(false)
```

and assert some property that should be updated (or not) in the update method. To switch the context, you can also use `tool.changeContext`

btw: for me, the code in ` oolComponent#update` looks very questionable (we wait for 500ms – why?). Later we run some code in the UI thread. There are definitely not enough comments.

HeikoBecker:
@alexanderpann :

@mgronover and I have looked at the test/code in a video call to clear up some questions regarding the `update` function. We have rolled back the change to `myNextContext` since it is not thread-safe. The use of `context` should not be removed.

There was a suppressed null pointer warning for the `context` object in the function. We have added a null-check and removed the suppressed warning. Since it has not yet occurred in practice, this change should be fine. Any concerns from your end?

HeikoBecker (approved):
LGTM

Issue #2309, opened 2023-10-06
Doc aspect: Allow more control about which concept docs are shown for a node


kbirken:
The documentation aspect allows [showing concept-specific documentation when the cursor in an editor hits a node of that concept](http://mbeddr.com/mps-platform-docs/aspects/documentation/?h=document), see `DocumentedConceptAnnotation`.

I am now looking for a configurable way to show this concept-specific documentation also if the cursor is on a child node (even when we have to override the concept-specific documentation of the child node).

This would allow a much better user experience in some situations. In the following example, the user has to click on the “:” specifically to see the documentation for the concept represented by the green line. All the documentation shown when clicking on some other green text would be much too specific.

<img width="299" alt="image" src="https://github.com/mbeddr/mbeddr.core/assets/1193045/d78be130-bab5-4c39-ae51-c1eea34febb2">

PR #2313, opened 2023-10-12
Allow override logic for concept-specific documentation in mbeddr.doc.aspect

kbirken:
This PR introduces a new flag `overrideChildren` for `DocumentedConceptAnnotation` nodes. Setting this flag will force the document for this concept be shown even if the cursor is not directly on a node of this concept, but on some child node (transitively). So the documentation for the child node will not be shown if there is an ancestor with the `overrideChildren` flag switched on. 

In addition to the feature, there are some more additions related to the feature:
- The configuration extension point `documentationAspectConfiguration` has been extended by the API call `allowOverrideChildren()` and documented. The `override`-logic is switched off by default, in order to use it, an extension has to be provided which returns true for `allowOverrideChildren()`.
- Example language:
  - A concept `CostAttribute` has been added to the example language in order to allow testing the new feature. An example node with an override-flag can be found here: [node URL](http://127.0.0.1:63320/node?ref=r%3A9f051296-ca8b-43ff-8d91-e0b7ca63b52a%28com.mbeddr.doc.lang.exampleLanguage.sandbox%29%2F2265458908607261733).
  - How to test? Open the example node. Open the documentation view Click on the expression nodes behind "cost: .." in the example, still the concept documentation for `CostAttribute` will be shown (see screenshot below).
  - An extension has been implemented for the default language to switch on `allowOverrideChildren()`.
- A cache for concept-specific documentation has been implemented. This improves the current performance (without the override-feature), but is really necessary when the override-feature is switched on.
- Introduce a blacklist for baselanguage root nodes. This avoids that the performance-heavy override-logic will be implied for any baselanguage code. Without this, the first click (with a cache miss) into some base language code node might take about 500-1000 milliseconds (from the second click on, there is no more performance penalty).
- The CHANGELOG has been updated.
 
Performance remark: I tested the override-feature in a big customer project. For an average root node for some arbitrary customer DSL, the first click with opened documentation view could take 50-300 milliseconds. After that, the cache is armed and the penalty is below 1 millisecond.

This PR solves #2309.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,15 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 ### Changed
 
-Added possiblity to update the ToolWindow contents of context action 2 evenif the Window is not visible.
+Added possiblity to update the ToolWindow contents of context action 2 even if the Window is not visible.
+
+## com.mbeddr.doc.aspect
+
+### Added
+
+* For documentation annotations of concepts there is a new flag `override children` in the inspector. It allows to show the document of a node even if the currently selected child node would have its own documentation.
+* By default, this new behavior is switched off. The `documentationAspectConfiguration` extension point now has a configuration option `allowOverrideChildren` which has to be set to true in order to use the override functionality.
+* A cache has been added to speed up the look-up and display of concept-specific documentation in the documentation view.
 
 # September 2023
 
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -19085,6 +19085,11 @@
             </node>
           </node>
         </node>
+        <node concept="1SiIV0" id="3KxJFmNUEug" role="3bR37C">
+          <node concept="3bR9La" id="3KxJFmNUEuh" role="1SiIV1">
+            <ref role="3bR37D" to="ffeo:mXGwHwhVPj" resolve="JDK" />
+          </node>
+        </node>
       </node>
       <node concept="1E1JtA" id="5NpY9mns5GC" role="2G$12L">
         <property role="BnDLt" value="true" />
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -19090,6 +19090,11 @@
             <ref role="3bR37D" to="ffeo:mXGwHwhVPj" resolve="JDK" />
           </node>
         </node>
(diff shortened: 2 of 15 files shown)


kbirken:
@wsafonov Added you as reviewer because we talked about how to avoid performance issues for this feature last week. Thanks for the fruitful discussion!

HeikoBecker (changes requested):
Changes look good overall. Here are some findings:

* Why does the demolanguage get so many new dependencies? This seems a bit off
* How about returning `false` for validation of cache entries in the case where the docNode is empty? This may lead to an additional overhead, yes, but it will make sure that new nodes are cached as well

alexanderpann:
This PR is open for 2 months. Can it be merged after those small modifications?

kbirken:
> This PR is open for 2 months. Can it be merged after those small modifications?

Thanks for the reminder. Up to now there were more high-prio request by the customers. But I'll address the review findings in the upcoming week and do the merge then.

@wsafonov Do you like to have a look on this PR before I merge it?

kbirken:
@HeikoBecker Thanks for your findings, I only now could find the time to look into it. But now the PR is ready for merge in my point of view, pls recheck.

> * Why does the demolanguage get so many new dependencies? This seems a bit off
Some of them are needed now because I implemented the extension point for the demo language. By running "Optimize Imports" on the language I could remove a few others which were actually not needed.

> * How about returning `false` for validation of cache entries in the case where the docNode is empty? This may lead to an additional overhead, yes, but it will make sure that new nodes are cached as well
Most of the concepts won't have any documentation, so whenever the user clicks on some node of one of these concepts, the doc aspect has to find out by a big search that there is no documentation. This is even more expensive when the override mechanism is activated and the whole chain of concepts from the clicked node to the containing root node is being processed. So to avoid this substantial additional overhead we have to cache the fact that a concept doesn't have a documentation aspect. Unfortunately we cannot decide to flush the cache in case a new documentation aspect for this concept is added - but this inconvenience only hits the documentation author, the bad performance hits every user.

A final remark: After our discussion about breaking changes in the platform group I also pushed a change in the extension point definition to avoid a breaking change for applications which implement the `IDocumentationAspectConfiguration` interface. See the javadoc comments for the interface and for `DefaultDocAspectConfiguration`. Maybe you could have a look into this when finalizing the review.

Thanks!

HeikoBecker (approved):
@kbirken Thanks for the explanations. They make sense to me.

I am approving now but would like @wsafonov to also have a look before this is merged.

alexanderpann:
Why is @wsafonov 's review needed here? I can also have a look if it is really necessary so that we can merge the PR.

kbirken:
> Why is @wsafonov 's review needed here? I can also have a look if it is really necessary so that we can merge the PR.

@alexanderpann I discussed this solution with him, but this is so long ago, that he probably forgot the details anyway. So please go ahead, I’m happy to respond to your questions and remarks. And thanks for bringing this up!

alexanderpann (approved):
LGTM. I also made a small performance improvement a short time ago in #2412 by not updating the documentation when the selection doesn't change. For new extension points in gernal, it would be nice to implement a `getPriority()` method where the extension with the highest priority is selected.

PR #2314, opened 2023-11-03
Fix NPE in BlockExpression constraints

sergej-koscejev:
Remove from BlockExpression_Constraints the call to `parentNode.isInTemplates()` whose return value was ignored but which caused a NPE if the parent node was null.


sergej-koscejev:
Superseded by #2315

PR #2316, opened 2023-11-08
Do not generate screenshots from MpsEnvironment, output a warning

sergej-koscejev:
CellScreenshooter now throws an error when it cannot generate a screenshot and we hit that exception when trying to generate screenshots from an Ant build. Instead of throwing an error, skip generating the screenshot and output a warning to point the user to a solution. Make sure to output the warning at most once per facet invocation, and only if there are screenshots.

PR #2317, opened 2023-11-08
Fix 'error when saving settings' caused by actionsfilter plugin

sergej-koscejev:
Includes #2316 (to avoid conflict in the changelog), please review that one first.

Passing `null` as action ID to IDEA's `CustomActionsSchema#addIconCustomization()` causes it to break later when persisting the settings. This didn't happen on 2021.3 so the fix is only needed for 2022.2 onwards.


HeikoBecker (approved):
Thanks for the fix!

Please only merge once the related PR (#2316) is merged.

sergej-koscejev:
Merging this anyway together with #2316 because we have no patience for false build problems.

PR #2320, opened 2023-11-09
Create a PR to merge a maintenance branch into the next one on push

sergej-koscejev:
When a maintenance branch pushed, create a PR to merge changes into the next version.

### DIFF
--- a/.github/workflows/PR-into-next-version.yml
+++ b/.github/workflows/PR-into-next-version.yml
@@ -1,0 +1,37 @@
+name: PR to merge into the next version
+
+on:
+  push:
+    branches:
+      - 'maintenance/*'
+
+jobs:
+  create-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Determine next branch and create PR
+        id: next-branch
+        shell: bash
+        run: |
+          # Next branch matching refs/heads/maintenance/* pattern, or empty if we're on the last one
+          next_maintenance_branch_full=$(git ls-remote --quiet --heads --sort=refname ${{ github.repositoryUrl }} 'refs/heads/maintenance/*' |
+            awk '$2 > "${{ github.ref }}" { print $2; exit }')
+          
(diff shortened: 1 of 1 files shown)


sergej-koscejev:
This action uses the maintenance branch itself for the PR to keep things simple. If we run into conflicts too often (for example, with the change log), we may have to switch to creating a merge branch first and using that.

.github/workflows/PR-into-next-version.yml
```diff
@@ -0,0 +1,37 @@
...
+
+          # Substitute the default branch if empty
+          target_branch=${next_maintenance_branch:-$default_branch}
+
+          echo "Target branch: $target_branch"
+
+          gh pr create --head ${{ github.ref_name }} --base ${target_branch} \
+            --title "Merge ${{ github.ref_name }} into ${target_branch}" \
+            --body 'This is an automatic PR which merges changes from `${{ github.ref_name }}` to `${target_branch}`.' \
+            --assignee '@me'
```
alexanderpann, line 37: Does that mean that the PR is assigned to some GitHub user/bot?

alexanderpann (approved):

.github/workflows/PR-into-next-version.yml
sergej-koscejev: I expect it to be assigned to the pusher who should be the one who merged the PR but I will need to try it out.

sergej-koscejev: It turns out it would be assigned to the bot, so I changed the logic to use the pusher explicitly as the assignee and the reviewer.

sergej-koscejev:
I have debugged the workflow [here](https://github.com/specificlanguages/automerge-sandbox), committing the final version. It will use the merge branch because I expect that conflicts will happen quite often because of the changelog, and also because we may want to add some sort of automated migration logic later.

Issue #2319, opened 2023-11-09
Fix and improve TraceExplorer


mgronover:
The TraceExplorer:
- is not always refreshing its content 
- should allow customers to select and use their own filters on the traced tree nodes
- should automatically update it's content when the filtering mode was change

PR #2328, opened 2023-11-22
[breaking] Bugfix/fix and improve trace explorer

mgronover:
This PR solves the issue: #2319

## Breaking Change:
The Interface of ITraceRecord was changed. It now requires a method:
`node<> getTargetNode()`;


alexanderpann (changes requested):
Looks good, just some small findings:

Typos:

- TraceFilterByITracerFrame#addThoseForFrames referncedNodes is misspelled
- TraceTab#Constructor selctionChangedListener is misspelled
- TraceTreeModel#allowedRepeatitions typo in comment
- Typo in "Choose one ore more options"

- TraceTreeModel#scanForChild is commented out. Not needed anymore?

UI:
- What about "Clear Editor from Tracing Information" - Remove any currently visible trace information within the code"?
- For all actions title case should be used e.g. "Next Trace for Same Node" instead of "Next Trace for same node"
- What about "Highlight in Editor" instead of "Highlight Code"?
- "Go to Computation Source" instead of "Goto computation source"
- What about "Reset Tree Expansions" instead of "Reset Tree"? For me, the button also doesn't seem to work or I don't understand it.
- What about "Repeat last Manual Unfolding" instead of "Unfold the tree as with last manual operation"?

- In RerunTraceAction there is an System.error statement which is also not visible to the user
- com.mbeddr.MPSUtils is misspelled in the CHANGELOG
- The rerun button doesn't seem to work:
I have the code:
```
fun times3(i: number) : number where pre i > 10                     = 3 * i
                                     post res > 100 && res == 3 * i        
```
And then an assert:
`assert times3(50) equals 200`

If I update the function to say `4 * i` instead of `3 * i` and I click rerun, shouldn't the trace be updated or do I not understand what the button does?

- The filter descriptions should also use title case per IntelliJ UI guidelines.
- "show nearly all trace nodes" is not very descriptive. What means "nearly all"?
- "show reduced trace tree (default)": the default string should be removed. Both filters are enabled by default.
- If you use a free-floating dialog for the filters, can you please open it relative to the position of the filter button? I am also pretty sure that there are real dropdown menus that you could use.

sergej-koscejev:
I don't quite like the added `getTargetNode()`. If this is only used for some sort of filtering, then maybe it should go into an extra interface, `IFilterableTraceRecord`? The filtering code checks for trace being `instanceof ComputationTrace` anyway, so why do we even need the extra method on the interface?

If we really need it, it should probably be a `SNodeReference` because it's not a good practice to hold on to nodes for a long time, especially across read/write actions (and yes, everybody is doing it but they really shouldn't).

Also, if we really need the method, perhaps it could have a default implementation just returning null.

alexanderpann (approved):
I am fine with the changes for now but I really think that the Rerun button is broken and It should work, for example, for assert items. We should really create a followup ticket for the other items.

PR #2332, opened 2023-12-06
Delete invalid assessment result entries

alexanderpann:
Automatically remove assessment result entries with invalid references e.g. a node or concept was removed. The automatic resolving of those references will fail and can completely lock up the editor if there are a lot of references.


mgronover (approved):
Well done.

PR #2344, opened 2023-12-14
Interpreter: generalize the evaluate methods into one method with options

alexanderpann:
The main motivation was to make the API easier and add support for disabling the computation trace.


mgronover (approved):
Looks fantastic!

Issue #2349, opened 2023-12-15
add a SPDX compliant license, organization and scm information into the created pom for a published mbeddr artifacts


KemalSoysal:
When checking for license policy constraints, the missing license information in the pom requires manual adjustment to the checking.

The pom.xml is described for the [license aspect](https://maven.apache.org/pom.html#licenses) and recommends to use the SPDX identifiers.

The [SPDX catalog](https://spdx.org/licenses/) states for the [Eclipse Public License V 1.0](https://www.eclipse.org/legal/epl-v10.html) the identifier `EPL-1.0`.

Add a license notice in the pom by creating the appropriate tags in the pom.xml

```xml
<project ...
...
  <licenses>
    <license>
      <name>EPL-1.0</name>
      <url>https://www.eclipse.org/legal/epl-v10.html</url>
      <distribution>repo</distribution>
      <comments>Eclipse Public License - v 1.0</comments>
    </license>
  </licenses>
...
</project>
```

Additionally Eclipse Public License V 1.0 requires the copyright notice. For this it could be enough to add the organization tag to the pom.xml

```xml
  <organization>
    <name>itemis AG</name>
    <url>https://www.itemis.com</url>
  </organization>
```

Additionally Eclipse Public License V 1.0 requires also the sources to be disclosed.
For accomplishing this requirement, provide the `scm` tag in the `pom.xml`. The tag is the commit hash just being built.

```xml
...
  <scm>
    <tag>${commithash}</tag>
    <url>https://github.com/mbeddr/mbeddr.core.git</url>
  </scm>
...
```

For the `graphviz` artifact, the license type is [Eclipse Public License Version 1.0](https://gitlab.com/graphviz/graphviz/-/blob/main/LICENSE?ref_type=heads) and has the identifier `EPL-1.0` and the scm repository is `https://gitlab.com/graphviz/graphviz`

PR #2350, opened 2023-12-16
#2349 add a spdx compliant license, organization and scm information to the created pom of published mbeddr artifacts

KemalSoysal:
Scanning projects for secure bill of materials (SBOM) is faciliated by the providing SPDX compliant
license, organization and source code information.

The modification on the build scripts enhance the created `pom.xml` by the information.

The users of the platform will benefit from the additional information by using automated processes that analyze the `pom.xml` in their projects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,26 @@ def findNonEmptyProperty(property) {
     return null
 }
 
+ext.additionalPomInfo = {
+    licenses {
+        // official SPDX identifier
+        // see https://spdx.org/licenses/ for list
+        license {
+            name = "EPL-1.0"
+            url = "https://www.eclipse.org/legal/epl-v10.html"
+            comments = "Eclipse Public License - v 1.0"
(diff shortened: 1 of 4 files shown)


alexanderpann (changes requested):
We are currently in the process of relicensing mbeddr based on customer requests. We probably switch to EPL 2.0 without a secondary license. Can you please use the 2.0 version? Thanks.

The alternative would be to not wait for the approval of the change and merge.

KemalSoysal:
> We are currently in the process of relicensing mbeddr based on customer requests. We probably switch to EPL 2.0 without a secondary license. Can you please use the 2.0 version? Thanks.
> 
> Please wait with the merging.

When the decision has been made, a change can be introduced by cherry-picking the commit for `EPL-2.0`.
I will not delete the original branch.

alexanderpann (approved):

PR #2356, opened 2024-01-03
Use reusable cascading merge workflow

sergej-koscejev:
Use reusable workflow from specificlanguages/cascading-merge


sergej-koscejev:
We may need to enable usage of external workflows for the mbeddr organization.

alexanderpann:
> We may need to enable usage of external workflows for the mbeddr organization.

We already have the following checkbox enabled: Allow all actions and reusable workflows

PR #2359, opened 2024-01-11
mpsutil.interpreter: Add a dummy computation trace when the computation trace is not available

alexanderpann:
A dummy computation trace is now used when the computation trace is not available in the interpreter to prevent NullPointerExceptions. This can happen when the interpreter is called with tracing disabled and the `trace` expression is used.


mgronover (approved):
LGTM

PR #2380, opened 2024-02-23
Use Sync instead of Copy in Gradle scripts

alexanderpann:


alexanderpann:
Does not work, that's why I close the PR.

Issue #2386, opened 2024-03-14
State charts visualization is broken


alexanderpann:
<img width="574" alt="Screenshot 2024-03-14 at 09 46 35" src="https://github.com/mbeddr/mbeddr.core/assets/88385944/5d7b127d-028b-48ed-bb73-0ca9a88c2127">

PR #2390, opened 2024-03-20
Documentation improvements

alexanderpann:
- also fixes https://github.com/mbeddr/mbeddr.core/issues/2386

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 ### Fixed
 
 - Many usability were fixed and the editors (including presentation mode) polished.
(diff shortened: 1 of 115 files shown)


HeikoBecker (commented):
Thanks a lot for the improvements. 

One small comment from my end:
Why is this model added? `Aaaaaa` looks strange
image

alexanderpann:
> Why is this model added?

It was wondered as well but it is just a test preferences page with an unfortunate name.

HeikoBecker (approved):
As we agreed, this is good to be merged once the `Aaaa` name is changed.

alexanderpann:
I'll make the name change in maintenance/mps20213.

PR #2394, opened 2024-03-28
markdown/plantuml: removed not needed dep. to mpsutil.httpsupport

arimer:
That dep. is not used and not needed and is causing dep. errors when those languages are packaged as part of an RCP.


alexanderpann (approved):
LGTM but maintenance/mps20213 would be the right branch for this bugfix.

arimer:
yeah, might be. I'll check

arimer:
closing in favor of https://github.com/mbeddr/mbeddr.core/pull/2395

PR #2221, opened 2022-06-23
seca-19.3 to maintenance mps 2020.3

heikob2:
Implement remaining changes from `maintenance/mps20193-seca` (not merged PR #2198) on top of `maintenance/mps20203` by hand using MPS version 2020.3.

### Affected
project mpsutil:
- asynccell changes (cherry-picked): updated
- editingGuide: `PropgramFragment` and `ExerciseExecutor`: minor changes
- hyperlink: `HyperlinkUtil`: minor changes
- actionsfilter: minor changes and some error fixes

project mbeddr.core
- `IMbeddrIDERoot` and `MbeddrMenuHelper`

### Changes not transferred
- project build: seems to be up-to-date
- graphstream: seems outdated
- incrementalcomputation: seems outdated
- AssessmentQuery: irrelevant

PR #2402, opened 2024-04-05
Integrate changes needed for SECURE

AlexeiQ:
This PR integrates necessary changes for itemis SECURE which were implemented on #2221 and therefore by #2198 by others. I've cherrypicked and updated the necessary commits and ommited all unnecessary/ now irrelevant changes after reviewing all the commits.

❗ The only change where I'm unsure and would like some advice/ additional review is: 
`com.mbeddr.mpsutil.actionsfilter.runtime`: Set the solution kind to `NONE`. (`eee1745d`)

## Changes
- `Force Save All` and `Execute Re-Runnable Migrations` on `com.mbeddr.core` and `com.mbeddr.mpsutil`

### com.mbeddr.mpsutil.editingGuide
- `ProgramFragments` support error suppression, i.e. all model-checking errors are suppressed in `ProgramFragments`, so modelchecker will disregard them and not clutter.
suppressedErrors
- Fragment module depends on its original module, to access data that is needed, but in the original module and not copied to the temporary exercise model (SECA-867).

### mpsutil.hyperlink.runtime
- `HyperlinkUtil.openInBrowser()` behaves like no-op instead of raising an exception on null/empty URLs.

### com.mbeddr.mpsutil.asynccell 
- Stop polling for the calculated value in case the poll-request throws an exception and show to the user, that the calculation has failed together with the thrown exception, instead of spamming the log on each poll with the exception.

### com.mbeddr.mpsutil.actionsfilter.runtime
- Set the solution kind to `NONE`. The actual reason for this change is not documented unfortunatly. It was probably either a simple clean-up or it caused class-loading issues in a project and therefore was necessary, which at the current time cannot be reproduced/verified anymore.

### com.mbeddr.core.base + .../pluginSolution
- Add possibility to explicitly not display roots of concept `IMbeddrIDERoot` in mbeddr menus.

### com.mbeddr.mpsutil.actionsfilter
- Implement missing behavior methods.
- Add error handling for missing actions.

## Checklist for creating a good PR (can be deleted)

- [x] I've described my change in the CHANGELOG.md if it is visible to users of the platform (**mandatory**).
- [x] I've selected the correct [base branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request).
- [x] I have checked my code (edge cases, no errors) and corrected any misspellings.
- [x] I have added comments to my code in areas where the code is hard to understand.
- [ ] I've created a new language/solution.
    - [ ] I've added the language/solution to the build script ([tutorial in readme](https://github.com/mbeddr/mbeddr.core#contributing-to-the-mbeddrplatform))
    - [ ] I've added a sandbox that showcases the language.
    - [ ] There is some kind of documentation available in the repository or somewhere else [e.g. [MPS Platform Docs](http://mbeddr.com/mps-platform-docs/platform_essentials/com_mbeddr_mpsutil/)].
 - [ ] I've introduced a breaking change.
     - [ ] I've made it clear that a change is breaking e.g. [breaking] prefix in the title
     - [ ] I have written new tests/changed old tests.
     - [ ] I have created migrations when necessary (e.g., deprecation of concepts, properties, enumeration values).
     - [ ] I haven't removed concepts, properties, references, or children without deprecating them first.
     - [ ] I have moved nodes or models by using the refactoring move action.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). The project does _not_ follow
 Semantic Versioning and the changes are simply documented in reverse chronological order, grouped by calendar month.
 
+# April 2024
+
+## com.mbeddr.core
+
+### Added
+- Base: Added possibility to explicitly not display roots of concept `IMbeddrIDERoot` in mbeddr menus.
+
+## com.mbeddr.mpsutil
+
+### Fixed
+- EditingGuide: `ProgramFragments` support error suppression, i.e. all model-checking errors are suppressed in `ProgramFragments`, so modelchecker will disregard them and not clutter.
+- EditingGuide: Fragment module depends on its original module, to access data that is needed, but in the original module and not copied to the temporary exercise model.
+- Hyperlink: `HyperlinkUtil.openInBrowser()` behaves like no-op instead of raising an exception on null or empty URLs.
+- Asynccell: Stops polling for the calculated value in case the poll-request throws an exception and shows to the user, that the calculation has failed together with the thrown exception, instead of spamming the log on each poll with the exception.
+
 # March 2024
 
 ## com.mbeddr.mpsutils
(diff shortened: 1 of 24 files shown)


alexanderpann (approved):
LGTM.

> com.mbeddr.mpsutil.actionsfilter.runtime: Set the solution kind to NONE  
>  

The change is probably fine because there is no plugin model in this module. Still, such a commit should be explained by the original author. I assume you tested this change.

AlexeiQ:
Thanks for the review and feedback :) 

> I assume you tested this change.
No, I did test everything else but the actionsfilter yet. I executed the gradle-command "behind" `./gradlew install`, but it didn't rebuilt the module, which is then copied to the MPS plugins directory. I have to do this tomorrow again properly.

AlexeiQ:
I did test the actionsfilter plugin by now, by calling `./gradlew install` and thus installing a new actionsfilter plugin into my MPS (cleaning the build and the plugin-directory beforehand). The actionsfilter is still working properly afai can see and the `idea.log` is free of any new, releated exceptions. After a discussion with @alexanderpann we agreed to `go on` with the merge.

Issue #2424, opened 2024-04-29
Provide helper classes for lazy execution of expensive operations


kbirken:
For expensive operations in baselanguage code, it is often required (due to performance reasons) to delay expensive operations until the result is actually needed for the first time. There is a pattern for this using the JDK `Supplier<T>` interface. We should provide this infrastructure in `mpsutil` so we can use it in other platform components of mbeddr or in IETS3.

kbirken:
PR is merged to maintenance/mps20222, closing ticket.

PR #2426, opened 2024-04-30
Add two memoizing supplier implementations

kbirken:
This PR provides two helper classes implementing the `Supplier<T>` interface. They can be used for lazy evaluation as well as lazy evaluation with an expensive and a cheap computation part. 

The PR also provides tests for these helper classes.

I also added existing tests for `c.m.mpsutil.compare` to the CI testing job, as I believe the tests are useful and should be executed on CI (boy scout's rule).

The change is not visible to users, so no CHANGELOG entry.

Solves #2424.


kbirken:
@dbinkele You wrote the original code of these classes, so maybe you want to check.
@alexanderpann Added as reviewer because I changed build scripts etc.

alexanderpann (approved):
LGTM, only the comments in the tests are lying about the actual value (777 vs 77).

kbirken:
> LGTM, only the comments in the tests are lying about the actual value (777 vs 77).

Thanks, fixed the test comments.

dbinkele:
@kbirken the tests cannot be run in the IDE: https://github.com/mbeddr/mbeddr.core/pull/2426

kbirken:
> @kbirken the tests cannot be run in the IDE: #2426

Works here. 

image

You probably have to uncheck "Execute in the same process":

image

Issue #2406, opened 2024-04-16
IllegalStateException: You must not register toolwindow programmatically so early.


mgronover:
```
Exception on a tool init: Tool com.mbeddr.doc.aspect.ui.plugin.TabbedDocViewer_Tool

java.lang.IllegalStateException: You must not register toolwindow programmatically so early. Rework code or use ToolWindowManager.invokeLater
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.getToolWindowPaneIfInitialized(ToolWindowManagerImpl.kt:500)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.registerToolWindow(ToolWindowManagerImpl.kt:930)
	at com.intellij.openapi.wm.ToolWindowManager.registerToolWindow(ToolWindowManager.kt:96)
	at jetbrains.mps.ide.tools.BaseTool.register(BaseTool.java:268)
	at jetbrains.mps.ide.tools.BaseTool.getToolWindow(BaseTool.java:198)
	at jetbrains.mps.ide.tools.BaseTool.toolIsOpened(BaseTool.java:101)
	at com.mbeddr.doc.aspect.ui.plugin.TabbedDocViewer_Tool.initializeTabsForSelection(TabbedDocViewer_Tool.java:83)
	at com.mbeddr.doc.aspect.ui.plugin.TabbedDocViewer_Tool.init(TabbedDocViewer_Tool.java:68)
	at jetbrains.mps.plugins.projectplugins.BaseProjectPlugin.registerPrefsAndTools(BaseProjectPlugin.java:130)
	at jetbrains.mps.plugins.projectplugins.BaseProjectPlugin.init(BaseProjectPlugin.java:97)
	at jetbrains.mps.plugins.projectplugins.ProjectPluginManager.createPlugin(ProjectPluginManager.java:159)
	at jetbrains.mps.plugins.projectplugins.ProjectPluginManager.createPlugin(ProjectPluginManager.java:54)
	at jetbrains.mps.plugins.BasePluginManager.createPluginChecked(BasePluginManager.java:153)
	at jetbrains.mps.plugins.BasePluginManager.createPlugins(BasePluginManager.java:143)
	at jetbrains.mps.plugins.BasePluginManager.loadPlugins(BasePluginManager.java:73)
	at jetbrains.mps.plugins.PluginLoaderRegistry.loadContributors(PluginLoaderRegistry.java:224)
[...]

PR #2434, opened 2024-05-08
Bugfix/tool window illegal state exception

mgronover:
This PR addresses issue  #2406


alexanderpann (changes requested):
I get the exception: jetbrains.mps.smodel.IllegalModelAccessError: You can read model only inside read actions

You probably need to surround the code inside ToolWindowManager#getInstance#invokeLater with command with repository { ... }

What also bothers me is that I don't see the documentation tool at the bottom of the screen when I don't use the show documentation action. Isn't this was `TabbedDocViewer#isAvailableOnStartup` should accomplish?

mgronover:
Thanks for the fix, @alexanderpann.

The visibility of the documentation tool at the bottom of the screen occurs sometimes, but is another point which could be handled in another ticket if really required.

alexanderpann (approved):

Issue #2435, opened 2024-05-15
Provide helper class for breadth-first traversals with cycle check


kbirken:
We are using breadth-first-traversals of structures (ASTs, POJO graphs, ...) in various locations of our code. A configurable, flexible traversal helper should be provided which can be used in those places instead of re-implementing the worklist/while/visited-algorithm again and again.

The helper should provide two APIs:

- an API with static methods for doing simple traversals as "one liners"
- an API as abstract class for doing more complex traversals

kbirken:
PR has been merged successfully, closing.

PR #2436, opened 2024-05-15
Provide utility class for breadth-first traversal and test suite

kbirken:
This PR contains the actual implementation `Traversal<T>` with a lot of javadoc documentation and many test cases. 

See the tests for examples how to use it. There are two different APIs which can be used:
- a compact API based on static methods and lambdas (used for simple traversals and searching for some specific node)
- a inheritance-based API with `@Override`, this can be used if additional data has to be stored and collect during the traversal

This solves #2435.

Here is an example of using the compact API from IETS3.Core:
image

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -22830,11 +22830,6 @@
             <ref role="3bR37D" node="5fGcQI947Ca" resolve="com.mbeddr.mpsutil.common" />
           </node>
         </node>
-        <node concept="1SiIV0" id="9jWrhFpXqx" role="3bR37C">
-          <node concept="3bR9La" id="9jWrhFpXqy" role="1SiIV1">
-            <ref role="3bR37D" to="90a9:PE3B26QCrP" resolve="org.apache.commons" />
-          </node>
-        </node>
         <node concept="1BupzO" id="9jWrhFpXqP" role="3bR31x">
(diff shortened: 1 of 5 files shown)


alexanderpann (approved):
LGTM. The tests also look correct to me. Please just add an entry to the README.

kbirken:
> LGTM. The tests also look correct to me. Please just add an entry to the README.

Thanks - which README do you think of? The one in the root folder doesn't talk about single classes or mpsutils content...

alexanderpann:
I meant CHANGELOG.md.

dbinkele:
@kbirken I have tried to improve the test assertions. I am really struggling with the testing of 'nVisited' as it tests the current traversal order of the nodes in the graph. To me this are to strict assumptions on the inner workings of the algorithm. This order especially depends on the order of the start vertices. So, starting with {2,3} can create different 'nVisited' than {3,2}.

kbirken:
> @kbirken I have tried to improve the test assertions. I am really struggling with the testing of 'nVisited' as it tests the current traversal order of the nodes in the graph. To me this are to strict assumptions on the inner workings of the algorithm. This order especially depends on the order of the start vertices. So, starting with {2,3} can create different 'nVisited' than {3,2}. I added ToDos in the Code highlightning 2 cases.

According to your todos: I think the order is guaranteed, for a given starting order if there is more than one start node. It is defined by the breadth-first order.  With a different starting order there can be a different traversal order, but this will again be deterministic. So from my point of view, the todos can be removed.

BTW: I replaced the type of the visited-collection to `linked_hashset` now, so that the `getVisited()` method returns the order of traversal and not an arbitrary one.

dbinkele (approved):
LGTM

Issue #2445, opened 2024-06-18
Code generation of pointer to ArrayType


mgronover:
There is a missing brace in the code generation of a pointer to an ArrayType.
And also some tests are missing for that.

[Here](http://127.0.0.1:63320/node?ref=r%3A0c020a2a-0fae-4359-90ba-63fdf45caa94%28tests.ts.core.tests1%40tests%29%2F722639934522900176) is the test case with the correct results within the comment below:
![image](https://github.com/mbeddr/mbeddr.core/assets/38247916/93d968fe-131b-4a27-9758-9d9ddd009993)

PR #2446, opened 2024-06-18
Bugfix/pointer to arraytype

mgronover:
This PR addresses issue: #2445


alexanderpann (changes requested):
The expected and actual values are switched in the test. The rest looks fine but looking through the values of TypeKind, I feel you also need parenthesis for FunctionRefType which is a function reference. We can keep it as it is but maybe you could check if my assumption is correct.

mgronover:
Thanks for the review!
I added also a test case for function pointers.

alexanderpann (approved):
LGTM. Please also update the CHANGELOG.

PR #2477, opened 2024-08-17
Remove platform distribution build script

sergej-koscejev:
It shouldn't be used because AFAIK we don't publish the `platform-distribution.zip` that it produces.


sergej-koscejev:
This is actually not true, it's used by the GitHub release.

PR #2480, opened 2024-08-19
mpsutils: create a new language for check-in handlers

alexanderpann:
Create a new extension for CheckinExtensionPoint to add a new handler. You can find some example handlers in `com.mbeddr.mpsutil.checkinHandler.demo.plugin`. The handlers can accept a commit, reject it or execute arbitrary actions.

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -20901,6 +20901,10 @@
         <ref role="m_rDy" node="3lcj7hzsuXf" resolve="com.mbeddr.mpsutil.logicalChild" />
         <node concept="pUk6x" id="3lcj7hzsE8a" role="pUk7w" />
       </node>
+      <node concept="m$_wl" id="SUkpD3RkHU" role="39821P">
+        <ref role="m_rDy" node="2hNr1jFzOYG" resolve="com.mbeddr.mpsutil.checkinHandler" />
+        <node concept="pUk6x" id="SUkpD3Rllc" role="pUk7w" />
+      </node>
     </node>
     <node concept="m$_wf" id="64SK4bcO2rO" role="3989C9">
       <property role="m$_wk" value="com.mbeddr.mpsutil.projectview.favourites" />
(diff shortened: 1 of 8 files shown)


sergej-koscejev:
I don't like the name `checkingHandler` because it looks to me like a misspelling of "check-in handler". Also, is it pre-commit handler or check-in handler? Let's use one term consistently as much as possible.

alexanderpann:
@sergej-koscejev That was not intentional, I changed the name to `com.mbeddr.mpsutil.checkinHandler`.

code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
```diff
@@ -247,6 +247,7 @@
   <node concept="1l3spW" id="3AVJcIMlF8l">
     <property role="TrG5h" value="com.mbeddr.platform" />
     <property role="2DA0ip" value="../../../../../build/com.mbeddr.platform" />
+    <uo k="s:LocalSourcePathArtifact" v="i:23468bc7" />
```
sergej-koscejev, line 250: Looks like some user objects were stored in the model? Where do they come from? Seems wrong.

alexanderpann:
Nice catch, I have fixed the build script but I don't know what happened.

code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
alexanderpann: Nice catch, I have fixed the build script but I don't know what happened.

sergej-koscejev:
The plugin should ideally only register a single handler on the IDEA side and internally this handler should walk through extension point objects every time it gets called. I'm worried that the current implementation will cause "module classloader is disposed" errors during development when modules with the check-in handlers are recompiled and reloaded.

But I'm going to approve this PR as it is a step forward.

sergej-koscejev (approved):

alexanderpann:
> module classloader is disposed

Yes, this is an issue, but shouldn't JetBrains try to find a fix? We have the same issue in other languages (e.g. iets3.opensource unit language) as you know.

sergej-koscejev:
This is not something they can fix, it’s basically a memory leak. Alternatively there should be a listener that would update the list of extension objects every time a module is loaded or unloaded.

PR #2493, opened 2024-09-01
Only `RunAntScript` tasks depend on `:downloadJbr`

sergej-koscejev:
Other tasks, such as `clean`, should not.


HeikoBecker (approved):
Thanks for the change! LGTM.

PR #2499, opened 2024-09-09
Checkin handler: add support for background checks + other small improvements and demos

alexanderpann:
There are now two extension points: `ForegroundCheckin` and `BackgroundCheckin`. The new background handlers support the interface `CommitCheck` which allows more options like running the code after the commit or in the background.


HeikoBecker (approved):
Thanks for this cool feature. LGTM

PR #2500, opened 2024-09-09
Build cleanup

sergej-koscejev:


build.gradle
```diff
@@ -86,10 +86,6 @@ ext.additionalPomInfo = {
 }
 
 subprojects {
-    // required plugins
-    apply plugin: "base"
```
alexanderpann, line 90: Does this mean that the message about local build detected, now also only shows once and not for every subproject?

subprojects/com.mbeddr/build.gradle
```diff
@@ -130,12 +135,6 @@ private static void configureRepositories(Project project) {
         }
     }
     project.repositories {
-        def disableMavenLocal = project.hasProperty("disableMavenLocal") && project.getProperty("disableMavenLocal").toBoolean()
-        if (!disableMavenLocal) {
-            // we don't use mavenLocal() repo, since it can cause various issues with resolving dependencies,
-            // see https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local
-            mavenLocal()
-        }
```
alexanderpann, line 138: I wonder if the reason for that flag was, that JetBrains needed it for their build. Before removing it, I would check this first but. We don't have it anymore in MPS-Extension though.

build.gradle
sergej-koscejev: I don’t think so, that’s future work.

subprojects/com.mbeddr/build.gradle
sergej-koscejev: They are not using it in the mbeddr build, see https://teamcity.jetbrains.com/buildConfiguration/MPS_20241_Distribution_TestMbeddrBuild?buildTypeTab=buildTypeSettings#all-projects

alexanderpann: This was the reason it was re-introduced: https://github.com/mbeddr/mbeddr.core/commit/85e3519b03ffe63c2520967525cb52a858539727

alexanderpann (approved):

subprojects/com.mbeddr/build.gradle
sergej-koscejev: Ah, okay, thanks, I’ll have another look at this.

sergej-koscejev: I removed the relevant commit as I want to make the build somehow do the right thing automatically without extra flags.

PR #2515, opened 2024-09-16
build.gradle: align snapshot publication with other platforms

arimer:
At the moment there are no snapshot artifact published for any branches except master/maintenance*/mps, which are published as regular artifacts.
To change this and to align the snapshot publication logic with our other platforms, the `mbeddrBuildNumber` is changed and prefixed with a `-SNAPSHOT`. 

In case that change works out, publication should go to the right default snapshot repository as defined in [here](https://github.com/mbeddr/mbeddr.core/blob/maintenance/mps20222/build/com.mbeddr/build.gradle#L110).


arimer:
@alexanderpann do you happen to know if any additional change needs to be done on the TC configuration side?

build.gradle
```diff
@@ -136,7 +135,13 @@ subprojects {
                 ext.mbeddrBuildCounter = GitBasedVersioning.getGitCommitCount()
             }
 
-            ext.mbeddrBuildNumber = GitBasedVersioning.getVersion(mbeddrBuild, mbeddrMajor, mbeddrMinor, mbeddrBuildCounter as int)
+            if(mbeddrBuild == "master" || mbeddrBuild.matches(/(maintenance|mps)[\/-](mps)?\d+(\.\d+)*/)) {
+                ext.mbeddrBuildNumber = GitBasedVersioning.getVersion(mbeddrBuild, mbeddrMajor, mbeddrMinor, mbeddrBuildCounter as int)
+            } else {
+                // use same logic as in all other platforms for snapshot publications
+                ext.mbeddrBuildNumber = GitBasedVersioning.getVersionWithCount(mbeddrMajor, mbeddrMinor, mbeddrBuildCounter as int) + "-SNAPSHOT"
```
sergej-koscejev, line 142: Shouldn't we add branch name to the snapshot version as well?

If not, then we don't need to use `getVersion` in the 'then' branch either, can use `getVersionWithCount` in both cases (and refactor the if).

arimer: You are right, I think a branch name should be part of the snapshot:

I'll add to this to keep that information.  
`GitBasedVersioning.getVersion(mbeddrBuild, mbeddrMajor, mbeddrMinor, mbeddrBuildCounter as int) + "-SNAPSHOT"`

However the snapshot publication is still not working. :(

sergej-koscejev: We need to add publishing to mbeddr / Pull Requests on TeamCity. But in order to do that we need to make sure it doesn't publish snapshots on GitHub.

arimer: Regarding the branch name agein: I Just checked _GitBasedVersioning.getVersionWithCount()_, and it should actaully determine the branch name [itself](https://github.com/mbeddr/mps-gradle-plugin/blob/master/src/main/groovy/de/itemis/mps/gradle/GitBasedVersioning.groovy#L77). However, it looks like it is not working properly for our TC config.

Issue #2527, opened 2024-09-23
Callback when adding trace nodes


mgronover:
For customer it is a hard job to figure out what the point of code created a certain node in the trace tree.
To answer this  question easily, we only have to introduce a call back function, called when a node is becoming a child of another one. 

(We have the situation, that a customer is not able to debug the source code.)

PR #2528, opened 2024-09-23
add callback function when trace node is add/inserted into the tree

mgronover:
addresses issue: #2527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 * Breaking change: debugger and spawner were not maintained and were removed.
 
+## com.mbeddr.mpsutil.interpreter
+
+### Feature
+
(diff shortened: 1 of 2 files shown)


alexanderpann (changes requested):
- Is it on purpose that you start iterating on this.parent and not this?
- Please add an entry to the README.
- Very minor but please use lowerCamelCase instead of UpperCamelCase for the static method names as it is normal in the Java world.

mgronover:
> * Is it on purpose that you start iterating on this.parent and not this?

You are right, if we start on iterating from `this` we also prevent inserting a node when `a.addChild(a)` is executed. So cycles on itself will be prevented. THANKS! 
I fixed this point.

> * Please add an entry to the README.

Done.

> * Very minor but please use lowerCamelCase instead of UpperCamelCase for the static method names as it is normal in the Java world.

Done.

PR #2532, opened 2024-09-23
Add graphstream language

sergej-koscejev:
A previously unpublished language for using the [GraphStream](https://graphstream-project.org/) graph library is now
part of the mbeddr platform distribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 * Breaking change: debugger and spawner were not maintained and were removed.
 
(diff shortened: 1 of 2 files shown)


alexanderpann (changes requested):
Let's keep this PR open, but after looking at the language, it has a few issues that need to be fixed:

- refreshing issues when hovering with the mouse over the tool (some elements disappear)
- a second copy of mxgraph that already exists in the diagramming language
- the runtime contains a library de.itemis.graphing.jar that we don't have the sources for
- the other stubs are so old that they have security vulnerabilities

sergej-koscejev:
Ok I will move the language into a separate build script so that it’s still not distributed but is at least being built.

PR #2548, opened 2024-10-07
`migrate` and `remigrate` tasks

sergej-koscejev:
New migration tasks, only migrating com.mbeddr.mpsutil and com.mbeddr.build for now, with more projects to follow.

Since migration will fail if an non-built language is encountered in a project, a new build script was added (sandboxes) to build all languages that were not built by other build scripts.

### DIFF
--- a/subprojects/com.mbeddr/build.gradle
+++ b/subprojects/com.mbeddr/build.gradle
@@ -160,14 +160,14 @@ private static void configureRepositories(Project project) {
 }
 
 
-task printVersions {
+tasks.register('printVersions') {
     doLast {
         println "mbeddrBuildNumber: $project.mbeddrBuildNumber"
         println "mbeddrPlatformBuildNumber: $project.mbeddrPlatformBuildNumber"
     }
 }
 
-task printRepositories {
+tasks.register('printRepositories') {
     doLast {
         println "snapshotRepository: $snapshotRepository"
         println "releaseRepository: $releaseRepository"
--- a/subprojects/com.mbeddr/build.gradle
+++ b/subprojects/com.mbeddr/build.gradle
@@ -1,3 +1,10 @@
+import de.itemis.mps.gradle.tasks.MpsMigrate
+import de.itemis.mps.gradle.tasks.Remigrate
+
+plugins {
+    id 'de.itemis.mps.gradle.common'
+}
+
 // path variables
 // If mpsHomeDir is set explicitly, skip the MPS resolution step and use the explicit path (which may be relative from
 // the root directory).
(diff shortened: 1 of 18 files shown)


subprojects/com.mbeddr/build.gradle
```diff
@@ -153,25 +160,59 @@ private static void configureRepositories(Project project) {
...
+}
+
+tasks.register('migrate', MpsMigrate) {
+    dependsOn(':com.mbeddr:platform:generate_platform_languages')
+    mpsHome = mpsHomeDir
+    projectDirectories.from(projectDirectoriesInDependencyOrder)
+
+    folderMacros['mbeddr.github.core.home'] = rootProject.layout.projectDirectory
+
+    pluginRoots.from(new File(mpsHomeDir, 'plugins'))
```
alexanderpann, line 200: Can we always do it like this instead of listing individual plugins? I don't care about the additional few seconds startup time.

subprojects/com.mbeddr/build.gradle
```diff
@@ -153,25 +160,59 @@ private static void configureRepositories(Project project) {
...
+tasks.register('remigrate', Remigrate) {
+    dependsOn(':com.mbeddr:platform:generate_platform_languages')
+    mpsHome = mpsHomeDir
+    projectDirectories.from(projectDirectoriesInDependencyOrder)
+
+    folderMacros['mbeddr.github.core.home'] = rootProject.layout.projectDirectory
+
+    pluginRoots.from(new File(mpsHomeDir, 'plugins'))
+
+    maxHeapSize = "4G"
```
alexanderpann, line 214: Is 4 GB really enough for mbeddr? I am probably I bit to cautious but shouldn't we set it to 8G to be sure?

alexanderpann (approved):
Thanks for the changes, looks good to me! Just a few comments:

`platform.sl-all`: can we please not use Sascha's old script name anymore? 

What bothers me a bit about the migrate/remigrate tasks is the time:
migrate: 1.45 min
remigrate: 2.32 min

Would it be possible to reuse the first MPS instance so that we don't have the expensive startup cost for the second task?

subprojects/com.mbeddr/build.gradle
sergej-koscejev: Yes, if it wasn't enough the build would have failed.

subprojects/com.mbeddr/build.gradle
sergej-koscejev: Where else?

sergej-koscejev:
> `platform.sl-all`: can we please not use Sascha's old script name anymore?

I don't see where I touched it - we should remove it, yes, but I'd do it in a separate PR then.

> What bothers me a bit about the migrate/remigrate tasks is the time: migrate: 1.45 min remigrate: 2.32 min
> 
> Would it be possible to reuse the first MPS instance so that we don't have the expensive startup cost for the second task?

`migrate` runs via an Ant task, `remigrate` uses a custom backend, so at the moment it's not possible to reuse the MPS instance. In order to reuse it, migration would have to be implemented in a backend as well.

sergej-koscejev:
I think I found the sl-all I introduced - in the sandboxes script. Removed it.

PR #2551, opened 2024-10-09
Upgrade to MPS 2023.2.1

sergej-koscejev:
2023.2 does not load project libraries in the IDEA environment (which is important for our migration tasks), 2023.2.1 fixes that.


alexanderpann:
Why not  2023.2.2?

sergej-koscejev:
> Why not 2023.2.2?

It's not published to our Nexus yet.

Issue #2547, opened 2024-10-07
Enable Caching of ILogicalChildren at ILogicalChildOwner


dbinkele:
Concepts 'ILogicalChildOwner' and 'ILogicalChildren' are designed such that the latter points to the former.
In case we need to find all 'ILogicalChildren' of a  'ILogicalChildrenOwner' currently the whole module is scanned everytime.

This needs to be improved by caching the result and updating on change without loosing the loose coupling described above

dbinkele:
done #2552

PR #2552, opened 2024-10-09
Add Caching mechanism (#2547)

dbinkele:
solves #2547


alexanderpann (changes requested):
Thanks for the addition! I just had a quick look: 

- You have to somehow derive the JBR executable from https://github.com/mbeddr/mbeddr.core/blob/ee1a4815061c27d3101bede860652e781f793058/build.gradle#L14 to not duplicate the logic.
- I really do not like seeing code for locking in our platform. Can you implemented is similar like [here](http://127.0.0.1:63320/node?ref=r%3A095345ad-6627-42ca-9d3f-fc1b2d9fbd61%28de.itemis.mps.editor.diagram.runtime.model%29%2F1993059304082831836) with a runnable? I am not 100% sure, but I believe marking the boolean variable as volatile should also be enough.

btw, you can use new singleton() from the collections language.

alexanderpann:
@dbinkele  I didn't understand your code with the locking at all, it doesn't look right to me.  I pushed a commit. Can you see if the change would be okay for you?

dbinkele:
> @dbinkele I didn't understand your code with the locking at all, it doesn't look right to me. I pushed a commit. Can you see if the change would be okay for you?

@alexanderpann This changes the semantics. The lock was there to make sure that Thread A is running in the 'withCachedLogicalChildren'-Environment and any Thread B who wants to do so as well must wait until A is finished. But both would do the caching. Now only does the caching, B starts  without waiting and probably does no caching ....

alexanderpann:
@dbinkele B does wait, I added another test case but I can't show a counter example where it doesn't work. The current thread blocks until it calls unlock. You can check the output of the second test.

alexanderpann (dismissed):

PR #2560, opened 2024-10-19
Simplify Gradle build scripts

sergej-koscejev:
Simplify the Gradle build scripts by removing support for overriding various settings via Gradle properties.

As far as I can tell, this support was only used by some special build running on GitLab and it is not running anymore.


sergej-koscejev:
> it would be nice to know why those customizations were necessary. My guess would be the old internal use of mbeddr at some of our locations.

@stephaneberle9 might know, he implemented some of these changes back in 2019 or so.

PR #2563, opened 2024-10-24
build: Use Gradle lifecycle tasks

sergej-koscejev:
The Gradle convention (as implemented by the built-in [`lifecycle-base` and `base` plugins](https://docs.gradle.org/current/userguide/base_plugin.html)) is to have a `build` task which runs `assemble` and `check`, where `assemble` builds the package and `check` runs tests (by depending on a `test` task) or any further checks such as linting, or in our case potentially model checks.

This PR implements this convention. As a prerequisite, it also decouples `package_*` tasks from `test_*` tasks so that tests only run when explicitly requested.

It should now be possible to simply run `build` or `check` to build and test the whole project.

### DIFF
--- a/build/com.mbeddr/languages/build.gradle
+++ b/build/com.mbeddr/languages/build.gradle
@@ -1,5 +1,9 @@
 import de.itemis.mps.gradle.*
 
+plugins {
+    id 'lifecycle-base'
+}
+
 // :com.mbeddr.build
 def script_build_mbeddr = new File(scriptsBasePath + "/com.mbeddr.build/" + "build.xml")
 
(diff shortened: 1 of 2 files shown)


alexanderpann (dismissed):
Thanks, the changes look reasonable to me.

sergej-koscejev:
For context, my goal here is to provide a simple interface between the build and TeamCity. Instead of listing all specific tasks to perform, TeamCity could be configured to only call `build` or `check`. This would make it possible to update build internals and rename tasks or refactor the subproject structure without having to adjust the CI build as much.

PR #2571, opened 2024-10-30
core.base: add back old condition for refreshing the tree view

alexanderpann:


sergej-koscejev:
Why change the migrate task configuration (cf9f479)? Was there a problem?

PR #2574, opened 2024-11-04
mbeddr.mpsutil: add a new language com.mbeddr.mpsutil.collections that provides an nset type

alexanderpann:
A new language `com.mbeddr.mpsutil.collections` was added that adds support for a set type `nset` that use nodes as the values of the set. Equivalence of nodes is checked structurally. The hash code calculation is done for all properties and children and the first level of references. The runtime solution also contains a more general class `EquivalenceHashSet` to implement hash sets with arbitrary `equals` and `hashcode` methods. `nset<>` is also compatible with `set<node>` and can be used as its type instead.

Implementation note: I didn't use the built-in implementation like the matches statement and the StructuralNodeSet from MPS because it deals poorly with copied nodes. For example, a copy of a class sometimes is not structural equal to the class itself: https://youtrack.jetbrains.com/issue/MPS-37894/matches-operation-not-working-for-type-variables

Screenshot 2024-11-04 at 19 54 33

Screenshot 2024-11-04 at 19 56 47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). The project does _not_ follow
 Semantic Versioning and the changes are simply documented in reverse chronological order, grouped by calendar month.
 
+# November 2024
+
+## com.mbeddr.mpsutil
+
+### Feature
+
+- A new language `com.mbeddr.mpsutil.collections` was added that adds support for a set type `nset` that use nodes as the values of the set. Equivalence of nodes is checked structurally. The hash code calculation is done for all properties and children and the first level of references. The runtime solution also contains a more general class `EquivalenceHashSet` to implement hashsets with arbitrary `equals` and `hashcode` methods.
+
 # October 2024
 
 ## com.mbeddr.mpsutil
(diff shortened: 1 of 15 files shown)


HeikoBecker (commented):
Thanks for the extension. This is really useful.

Here are some preliminary technical findings. I am not yet done with a full review:
- Do we need this constructor? It looks dangerous since it may change the behavior of the underlying set.
image
- Can you add a test that `NodeEquivalence.equals(a,b) -> NodeEquivalence.hash (a) == NodeEquivalence.hash (b)` for consistency of hashCode with equals? 
- This test seems duplicated:
image
- Nitpick, but I would prefer removing one element and testing for size equal 1:
image
- Why was `equals` replaced with `:eq:` in only this test: 
image

HeikoBecker (changes requested):
No other findings. Please do take a look at my initial findings. Sorry for splitting up the comments.

alexanderpann:
> Do we need this constructor?

That is the main point of the `NodeHashSet`class  to make the set behave differently for nodes. What do you suggest here?

I fixed all the other points.

HeikoBecker:
> > Do we need this constructor?
> 
> That is the main point of the `NodeHashSet`class to make the set behave differently for nodes. What do you suggest here?
> 
> I fixed all the other points.

I do not think that we need to remove it.
Can we maybe make it clearer that it is supposed to be a copy-constructor? This would in my opinion make it clearer that set properties may change.

HeikoBecker (approved):
Thanks for the changes. LGTM.

PR #2609, opened 2024-11-19
make sure mpsutil.compare.AreEqualExpression validates even without ignored properties or references

enikao:


alexanderpann (approved):
LGTM, thanks for the change again.

PR #2622, opened 2024-11-21
Fix exception on FindAction-action

enikao:
Prevent message when pressing Ctrl-Shift-A: "No display name specified in plugin descriptor XML file for configurable XXX; specify it using 'displayName' or 'key' attribute to avoid necessity to load the configurable class when Settings dialog is opened"

Happened for every PreferenceForm


alexanderpann (approved):
Thanks for the PR, I couldn't find a fix!

enikao:
> Thanks for the PR, I couldn't find a fix!

Yes, it was very indirect. Had to fiddle around a lot.

PR #2639, opened 2024-12-05
genfacade: use RegularPlanBuilder instead of RigidPlanBuilder

wsafonov:
Use RegularPlanBuilder instead of RigidPlanBuilder in [GeneratorFacade](http://127.0.0.1:63320/node?ref=r%3A00bd75cf-1225-4ef5-9a7e-390aed8718dd%28com.mbeddr.mpsutil.generatorfacade.runtime%29%2F5915735921188775088) to properly support enhanced Transform generation plan steps with multiple Transform.entries that should be executed in one single step (RigidPlanBuilder doesn't support this). Previously multiple languages specified in the Transform.languages were always combined in one generation step.


alexanderpann (approved):
LGTM. RigidPlanBuilder is even unused in MPS (except for 1 test).

PR #2660, opened 2024-12-17
use SoftMap instead of WeakMap to cache interpreters

enikao:
WeakReference is killed any time the garbage collector runs, whereas SoftReference is killed only if memory is low.

Should lead to more cached interpreters.


alexanderpann (approved):
LGTM, thanks for the change.

PR #2673, opened 2025-01-09
Don't use deprecated StyleRegistry singleton

artem-tikhomirov:
StyleRegistry.getInstance will be removed in MPS 2025.1


alexanderpann (commented):
While the colors nearly match with the light theme, they differentiate too much with the dark theme.

artem-tikhomirov:
Well, if there's intention for the field to match color with an editor, then I'd expect to see initial color being set all the time (w/o assuming light theme color generally matches that default of a text field). 
If there's no proper context for the editor in this case, perhaps, using `EditorColorsManager.getInstance().getGlobalScheme()` to get default theme foreground is preferable. Do you want me to update the fix?

alexanderpann:
Yes, please. Thanks.

alexanderpann (approved):
LGTM

PR #2674, opened 2025-01-09
[breaking] mpsutil.json: improve exporting of files and make parser and export options customizable through an extension point

alexanderpann:
The text generator output of `com.mbeddr.mpsutil.json` was improved and the escaping of special characters was implemented. The output options and JSON parsing can now also be configured through the extension point `json` in the method JsonConfig#getFactory. For more information read: https://github.com/fasterxml/jackson-core/wiki/JsonFactory-Features, https://github.com/fasterxml/jackson-core/wiki/JsonGenerator-Features and https://github.com/fasterxml/jackson-core/wiki/JsonParser-Features.

The PR is breaking since escaping characters can only be disabled for non-ASCII characters (through JsonGenerator.Feature.ESCAPE_NON_ASCII) and not for the quotation character which would produce an invalid JSON file. In case there is another change regarding the output, it should be possible to restore the old behavior through modifying the generator through the extension point.

While working on the PR, I noticed that we can't parse large numbers with the Jackson library. They are parsed as "Infinity"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,13 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 # January 2025
 
-# December 2024
-
 ## com.mbeddr.mpsutil
 
 ### Fixed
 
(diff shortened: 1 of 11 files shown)


HeikoBecker (dismissed):
Thank you for the change. This is much appreciated.

PR #2675, opened 2025-01-12
Reflect j.m.refactoring.participant is a solution now

artem-tikhomirov:
j.m.refactoring.participant used to be a language with sole 'plugin' aspect, now converted to a solution, and a reference in the build script needs to reflect new node


sergej-koscejev:
@artem-tikhomirov I cherry-picked your commit to #2649 but now I get errors related to CopyPasteUtil so there are probably more changes needed around copy-paste in order to compile against the latest MPS build. It would be best if you could make them on the branch renovate/mpsbuild.

sergej-koscejev:
I think I made the same fixes on renovate/mpsbuild and I am doing a few more fixes to make it green there.

PR #2678, opened 2025-01-13
build: upgrade to Gradle 8.12, introduce jbr-toolchain

sergej-koscejev:
Switch from `download-jbr` to [`jbr-toolchain`](https://github.com/specificlanguages/mps-gradle-plugin/tree/d0783ed6c99d6a3bb9e4e9a29fcec679ea4e8550/subprojects/jbr-toolchain).

The JBR version is now moved to version catalog where it is more visible to Gradle tooling, e.g. it can be upgraded by Renovate and/or locked.

Using jbr-toolchain also allows sharing the extracted JBR among multiple projects.

This PR also introduces build-logic (written in Kotlin) as a place where boilerplate code can be offloaded from the build scripts, and upgrades Gradle to 8.12.


sergej-koscejev:
This is part of my effort to better separate mbeddr core from mbeddr C. In order to achieve that, it should be possible to specify dependencies such as the JBR, MPS, and MPS-extensions between the projects as Gradle dependencies, rather than by depending on some tasks in some projects that unpack files into a fixed location.

alexanderpann (approved):

PR #2684, opened 2025-01-15
Reflect MPS master changes (buildscript and clipboard api)

artem-tikhomirov:
This branch contains two fixes, and obsoletes https://github.com/mbeddr/mbeddr.core/pull/2675 request.


sergej-koscejev:
I think this PR is no longer needed, it does not contain any significant changes anymore.

Issue #2670, opened 2024-12-20
Deployed artifact contains unwanted stuff in icons folder


kbirken:
When mbeddr.platform is loaded into the `build/dependencies` folder by gradle, there is a subfolder `com.mbeddr.platform/mbeddr.mpsutil/icons` which contains stuff that should not be part of the delivery. Here are the examples which triggered me, but there might be more:

```
./mbeddr.mpsutil/icons/test.com.mbeddr.mpsutil.logicalChild
./mbeddr.mpsutil/icons/test.com.mbeddr.mpsutil.logicalChild/models/test.com.mbeddr.mpsutil.logicalChild.test@tests.mps
./mbeddr.mpsutil/icons/test.com.mbeddr.mpsutil.logicalChild/test.com.mbeddr.mpsutil.logicalChild.msd
./mbeddr.mpsutil/icons/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox
./mbeddr.mpsutil/icons/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox/models/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox.testbuffer.mps
./mbeddr.mpsutil/icons/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox/models/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox.example.mps
./mbeddr.mpsutil/icons/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox/com.mbeddr.mpsutil.logicalChild.sandbox.sandbox.msd
```

These shouldn't be part of the deployed artefacts at all. In this particular case, it leads to warnings in customer projects, e.g.

```
2024-12-20 15:57:20,326 [1688805]   WARN - jetbrains.mps.classloading.ModulesWatcher - com.mbeddr.mpsutil.logicalChild.sandbox.sandbox was marked invalid for class loading: The dependency in com.mbeddr.mpsutil.logicalChild.sandbox.sandbox [solution] cannot be resolved {Default} 286e2375-00e3-4042-b083-84873dd623be(com.mbeddr.mpsutil.logicalChild.sandbox)
2024-12-20 15:57:20,326 [1688805]   WARN - jetbrains.mps.classloading.ModulesWatcher - test.com.mbeddr.mpsutil.logicalChild was marked invalid for class loading: The dependency in test.com.mbeddr.mpsutil.logicalChild [solution] 
```

[...]

PR #2690, opened 2025-01-20
Ensure that packaged artifact does not contain unwanted solutions

kbirken:
This solves issue #2670.

Currently, the `com.mbeddr.platform` build script specifies a folder `icons` for the plugin `com.mbeddr.mpsutils`. This folder collects all contents from various `icons` folders in the project. However, it also listed the full directory `$mpsutils/solutions`, which leads to lot of stuff being contained in the deployed `com.mbeddr.platform` artifact. This also leads to some errors and warnings in the logs when MPS is running. 

This PR replaces the item `$mpsutils/solutions` by all folders below this directory named `icons` (which are actually only two). I also checked for files below this folder with extensions `*.png` and `*.svg`, but did not find other icon files outside of the two sub-folders I added.

Risk: There might be files in the packaged folder `$mpsutils/solutions` which are required downstream, but not packaged after this PR. I do not know how to check this.

One open topic: During my analysis I found the folder 

```
com.mbeddr.mpsutil/solutions/doc.com.mbeddr.mpsutil.multilingual/images
```

which contains a couple of png-images. We have to check if they are actually needed and if yes, if they are packaged to a different folder.

### DIFF
--- a/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
+++ b/code/languages/com.mbeddr.build/solutions/com.mbeddr.platform/models/com/mbeddr/platform/build.mps
@@ -20202,18 +20202,38 @@
                 <property role="2Ry0Am" value="languages" />
                 <node concept="2Ry0Ak" id="3TSnT3Ivcei" role="2Ry0An">
                   <property role="2Ry0Am" value="com.mbeddr.mpsutil.multilingual.editor" />
-                  <node concept="2Ry0Ak" id="3TSnT3Ivcej" role="2Ry0An">
+                  <node concept="2Ry0Ak" id="5jdrrsPOhHp" role="2Ry0An">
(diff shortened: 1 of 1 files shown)


arimer:
Jup, just saw the same issue in the artifacts. Starting from this artifact on, the icons folder was poluted: https://artifacts.itemis.cloud/#browse/browse:maven-mps:com%2Fmbeddr%2Fplatform%2F2023.2.25014.09358e0%2Fplatform-2023.2.25014.09358e0.zip

Looks like there was a change in this PR causing this: https://github.com/mbeddr/mbeddr.core/pull/2627

arimer (approved):
Thanks for cleaning up the build script. The changes look good to me. The generated artifact should now also be significantly smaller than before. 👍🏼

arimer:
I would opt for fixing future findings related to wrong/missing icons when they pop up. Having a cleaned up artifact which is used by customers is more important.

Issue #2696, opened 2025-01-21
Command line execution of interpreter test evaluation produces exceptions


mgronover:
When evaluating the interpreter tests from command line many exceptions like this one are produced:

`Argument for @NotNull parameter 'node' of org/jetbrains/mps/openapi/model/SNodeUtil.getDescendants must not be null`
`<...stack dump follows here...>`

mgronover:
This was solved [elsewhere](https://github.com/IETS3/iets3.opensource/issues/1144#issuecomment-2630173323)

PR #2697, opened 2025-01-21
enhanced type determination for interpreter constraints #2696

mgronover:
This PR addresses #2696.


arimer:
those changes in the build look odd. Do you mind reverting them? 
grafik

mgronover:
This is not required! It was created before understanding that devkits weren't loaded.

Issue #2052, opened 2019-03-09
Shrink the mbeddr platform


coolya:
Right now the mbeddr platform contains some code that is deprecated and largely unmaintained. Migration of these parts into MPS extensions is highly unlikely. We want to removed this code in the future (after the MPS 2019.1 release has stabilized) in order to lower our maintenance overhead  and to give a clear indication what the quality of the individual parts of the platform are. 

As an input for the discussion we will use the quality assessment done by @enikao, revisit some of the parts assessment result and then communicate a list of components we want to remove, migrate to MPS extensions or keep inside the platform. 

We invite the community to provide us feedback on out proposals so that we can take informed decision about the removal. 

So far two components are scheduled for removal:

- multilingual 
- breadcrumb (is disabled for at least two MPS releases and no complains where filed)

PR #2711, opened 2025-02-02
Removing com.mbeddr.mpsutil.multilingual.common support from behavior of com.mbeddr.core.base and cleanup imports.

malueck:
This PR removes the `com.mbeddr.mpsutil.multilingual.common` dependency from the behavior model of `com.mbeddr.core.base` (as well as from the `mpsutil.editingGuide` sandbox where it was an unused import).

The empty refactoring, actions and plugin models were also dropped from com.mbeddr.core.base and a few imports cleaned up.

The `IValueDebuggable_Behavior` node in the model had a single use of multilingual in the `getDebuggerValue()` method. Instead of returning the multilingual translation with a default string of `"(no value source)"`, the  English string "(no value source)" is now always returned by `getDebuggerValue()`.

(In the same node, also a deprecated instance of new Long() was removed, see https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Long.html#%3Cinit%3E(long) )

Justification:
- The plugin is scheduled for removal according to https://github.com/mbeddr/mbeddr.core/issues/2052, however, instead of removing it we at least exclude it from the mbeddr.core dependencies because there is still another remaining use in `mpsutil.preferenceform`.
- The `multilingual` plugin is also listed as deprecated in the MPS Platform Docs: http://mbeddr.com/mps-platform-docs/platform_essentials/mpsutils/deprecated/#multilingual
- This forms a dependency  that "leaks through" the build dependencies. `com.mbeddr.core.base` is part of the group, plugin and build script all named `com.mbeddr.platform`.
   - However, neither the plugin nor the group contain nor depend on `com.mbeddr.mpsutil` which contains `multilingual`. (It seems like the model checker notices missing dependencies in "Dependencies", but not in "Used Languages".)
   - This can break other plugins, such as the build script of `org.iets3.core.os` which has a build dependency on `com.mbeddr.platform` to get `com.mbeddr.core.base`, but not `com.mbeddr.mpsutil` to get  `com.mbeddr.mpsutil.multilingual.*`.
   - It looks like this will resolve the issue https://youtrack.jetbrains.com/issue/MPS-38155/Languages-from-plugin-are-not-acknowledged-as-deployed and https://github.com/IETS3/iets3.opensource/issues/1195


alexanderpann (approved):
LGTM. Thanks for the contribution and also the cleanup! I'll take care in case the iets3.opensource plugin breaks. Let's see if it helps with your loading issues or if further adjustments are needed.

alexanderpann:
@malueck can we close those 2 mentioned tickets and is the issue resolved for you?

malueck:
Thanks!
I opened a new issue in https://github.com/IETS3/iets3.opensource/issues/1207 (now hopefully at the correct place) about a few more wrong dependencies.
However, I am now again able to use the basic languages like `org.iets3.core.expr.base`. So, for me, the issue is resolved.

I would however like to keep at least the MPS Youtrack issue open, I will try to add a minimal example there for the case that we had, i.e. we have languages dependencies A -> B -> C, and MPS shows that A and C are not deployed due to dependency issues, but shows B as deployed, which is a bug in my opinion.

PR #2712, opened 2025-02-02
build: remove the 'allScripts' solution

sergej-koscejev:
* Generate build scripts using `MpsGenerate` task.
* Remove the corresponding Maven publication since it brings no value.


alexanderpann (approved):
LGTM, I don't see any references to com.mbeddr.allScripts from other projects

PR #2741, opened 2025-02-07
build: fix building with custom MPS

sergej-koscejev:
fileValue wants a File but mpsHomeDir is a String.


sergej-koscejev:
Bypassing rules and merging because our CI wouldn't test this properly anyway.

PR #2744, opened 2025-02-10
Merge maintenance/mps20243 into master

mps-ci-buildbot:
Merge changes from `maintenance/mps20243` to `master`.


sergej-koscejev:
On master the dependency on MPS-extensions is fixed, not range- or wildcard-based.

PR #2761, opened 2025-02-23
refactor: Separate platform build

sergej-koscejev:
Extract the build solutions of platform project into a separate MPS project, `com.mbeddr.platform.build`.

The change should not be visible to the users.


alexanderpann (commented):
The changes look correct, and it builds, but you might want to ask @wsafonov to get a better review because I am not that familiar with the Gradle build scripts.

wsafonov (approved):
Sorry for the delay. Generally looks good to me, can be merged. Just one question for my understanding: using evaluationDependsOn() seems like a hack for now, right? I guess it should be gone as soon as we get a more clearer project separation with build scripts per project.

PR #2766, opened 2025-02-26
bugfix(doc): improve memory usage of generateScreenshots facet

sergej-koscejev:
* MakeScreenshotRunnable does not need to hold on to its image result.
* Catch all Throwables while running in order to also catch OutOfMemoryErrors.


alexanderpann (approved):
LGTM

PR #2831, opened 2025-03-26
Include all MPS projects into automated migration

sergej-koscejev:
Two changes that may potentially break things:
* The move of spreadsheet libraries to a separate stub solution.
* The change in the doc language to only drop unused Document _roots_ during generation.


alexanderpann (dismissed):
LGTM. The stubs for the spreadsheet language are definitely a breaking change because the POI library is sometimes used by people to read or write Excel files.

Why was the change in the doc generator necessary?

alexanderpann (approved):

PR #2843, opened 2025-04-01
Bundled dependencies & platform SBOM publishing

wsafonov:
Introduced Gradle dependencies for all bundled library jars used in the `com.mbeddr:platform` languages package (that have been previously stored directly in the git repo) and added SBOM for the `com.mbeddr:platform` package to the published zip archive.

Some libraries have been updated to the next bugfix version. Here is an overview of the changes:

- `org.commonmark:commonmark` used in `com.mbeddr.doc.gen_markdown` updated from `0.17.1` to `0.17.2`
- `com.michaelbaranov:microba` used in `com.mbeddr.mpsutil.datepicker.runtime` updated from `0.4.4` to `0.4.4.3`
- `rhino-1.7.10.jar``removed from `com.mbeddr.mpsutil.plantuml` (as not used anymore)
- `au.com.bytecode:opencs` used in `com.opencsv` updated from `2.3` to `2.4`
- `xalan:serializer` used in `org.xml` updated from `2.7.1` to `2.7.3`
  `xerces:xercesImpl` used in `org.xml` updated from `2.11.0` to `2.12.2`
  `xml-apis:xml-apis-ext` used in `org.xml` updated from `1.3` to `1.3.04`


subprojects/com.mbeddr/platform/build.gradle
```diff
@@ -2,13 +2,23 @@ plugins {
     id 'base'
     id 'maven-publish'
     id 'buildlogic.mps-conventions'
+    id "org.cyclonedx.bom" version "1.8.2"
```
alexanderpann, line 5: Why not the newest version (2.2.0)?

subprojects/com.mbeddr/platform/build.gradle
```diff
@@ -169,3 +249,14 @@ if({project.mbeddrBuild}() == "master") {
        */
     tasks.findByName("publishMbeddrPlatformPublicationToMavenRepository").dependsOn("publishMbeddrPlatformPublicationToGitHubPackagesRepository")
 }
+
+cyclonedxBom {
+    // SBOM destination directory
```
alexanderpann, line 254: Not sure if it is relevant, but I set schemaVersion = "1.5" to have the same version as in iets3.opensource.

alexanderpann:
Do we also want to update the dependencies block in https://github.com/mbeddr/mbeddr.core/blob/4ca76f70045a4e1278e0a73a66133c75b1d16150/subprojects/com.mbeddr/platform/build.gradle#L214? MPS-Extensions is not included but shouldn't those other external dependencies be declared here?

wsafonov:
> Do we also want to update the dependencies block in
> 
> https://github.com/mbeddr/mbeddr.core/blob/4ca76f70045a4e1278e0a73a66133c75b1d16150/subprojects/com.mbeddr/platform/build.gradle#L214
> 
> ? MPS-Extensions is not included but shouldn't those other external dependencies be declared here?

Provided dependencies have actually no real use case, in the past we just had MPS as provided dependency, but for no practical reason, just as additional information. So I would rather keep the bundled dependencies all out for now to not pollute the POM. The required information about used components will be provided in the SBOM anyway, so POM can be kept minimal for now.

PR #2904, opened 2025-04-16
Update bundled dependencies

wsafonov:
- `jcommon.jar` is not available anymore as transitive dependency with the recent used jfreechart version and has been therefore removed from the `c.m.mpsutil.jfreechart.runtime` solution
- `poi-ooxml` jars have been updated according to the transitive dependencies of the currently used version


kbirken (approved):
Reviewed together with @wsafonov. Looks good.

kbirken (approved):
Still fine :-)

PR #2914, opened 2025-04-28
Switch to batik provided by mps-extensions

arimer:
This PR backports multiple changes from maintenance/mps20241 to get rid of batik and switch to a batik version provided by mps-extensions.

This PR depends on the according mps-extensions [PR](https://github.com/JetBrains/MPS-extensions/pull/1328) that provides batik + xml stubs for MPS 2023.2. This PR needs to be merged first.


arimer:
@wsafonov , @sergej-koscejev any change one of you could have a look?

subprojects/com.mbeddr/platform/build.gradle
```diff
@@ -55,7 +55,6 @@ def bundledDeps = [
             exclude module: 'commons-compress'
             exclude group: 'org.bouncycastle'
             exclude module: 'xmlsec'
-            exclude module: 'batik-all'
```
wsafonov, line 58: Why did you remove this exclude for the transitive dependency? Batik stuff is not bundled and not used in apache-poi anyway. It actually won't change anything since the jars would just be downloaded, but not used. So for consistency I would prefer to also keep the dependency excluded here.

code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.graphstream/runtime/com.mbeddr.mpsutil.graphstream.runtime.msd
```diff
@@ -57,6 +57,7 @@
     <module reference="5787a8ed-1486-4469-94b0-fa3fc6c8538d(com.mbeddr.mpsutil.graphstream)" version="0" />
     <module reference="ab71436a-a7d1-4689-ac02-b5fde2ec681f(com.mbeddr.mpsutil.graphstream.runtime)" version="0" />
     <module reference="c0488c1e-322f-4f38-92d4-5520a7ce96c1(com.mbeddr.mpsutil.plantuml.pluginSolution)" version="0" />
+    <module reference="563a0770-eb88-4f4f-83ed-f708776fc2fe(de.itemis.stubs.batik)" version="0" />
     <module reference="5b1f863d-65a0-41a6-a801-33896be24202(jetbrains.mps.ide.editor)" version="0" />
     <module reference="ceab5195-25ea-4f22-9b92-103b95ca8c0c(jetbrains.mps.lang.core)" version="0" />
     <module reference="5ad14eca-28d7-4bce-b8e0-648908a49062(org.apache.batik)" version="0" />
```
wsafonov, line 63: Although the actual solution is removed now, there is still this version reference, which is weird. Maybe check if everything has been properly force-saved?

wsafonov (changes requested):
Have just couple of questions to clarify, see inline comments.

subprojects/com.mbeddr/platform/build.gradle
arimer: tbh. I just removed than, cause I though: mbeddr is no longer dealing with batik, so let's ditch it.

code/languages/com.mbeddr.mpsutil/languages/com.mbeddr.mpsutil.graphstream/runtime/com.mbeddr.mpsutil.graphstream.runtime.msd
arimer: Removed. It was part of the depdenecy version..

wsafonov (changes requested):
It looks like there are still some module references remaining in the msd files, at least I see `org.xml` and `org.apache.batik` in plantuml. Maybe there are more of these things left in other places. Actually I was assuming that force-save would resolve this issue...

arimer:
@wsafonov yeah, it looks like the `dependencyVersions` are not updated via the force-save all action :(

arimer:
@wsafonov I removed the remaning, outdated dependencyVersions manually and cleaned up the modules.xml as well. 
Fell free to have a look.

wsafonov (approved):
Looks clean now, thanks!

Issue #2911, opened 2025-04-28
Provide utilities for implementing instance tree path classes


kbirken:
# Problem description

In many projects classes are being implemented which represent a path in some (instantiation) tree structure. A _path_ consists of
- a _root_ object (optional), defining the starting point for the path (typically a reference to the root of the instantiation tree)
- a sequence of _segments_, defining the navigation along the nested tree structure

Often, each tree node (represented by a segment) refers to its definition object. These definition objects and  the root object have the same type, the _definition type_.

Typical operations include:
- constructing paths
- checking two paths for equality
- appending new segments to an existing path
- concatenating two paths (if they match)
- converting a path into some (dot) expression

# Example application

A typical component DSL:
```
component A
   instance B as b1
   instance B as b2

component B
    instance C as c1
    instance C as c2
    instance C as c3

component C
```

Valid pathes in this example would be `A.b1.c2` or `A.b2` or `A` (all of these have root `A`) or even `C` (when root is `C`).

# Solution

A set of utility classes should be provided which support the implementation of instance path classes as described above. The resulting path classes should be immutable, i.e., existing path objects are not changed, but new path objects are created if operations produce a changed path object.

PR #2916, opened 2025-04-29
Provide base classes for instance-path classes for nested structures

kbirken:
This solves #2911, see description of the functionality there.

The PR comes with an example application class (for base language), and provides a couple of tests. We intend to use the new helper classes for at least two applications in IETS3-Core (e.g., ArtifactPath), and possibly in some customer applications.


mfsoliveira (approved):
Nice support for our Instances!
The review points were discussed directly with Klaus.

mfsoliveira (approved):

PR #2923, opened 2025-04-30
MPS-38286 fix reported incomplete plugin dependencies

artem-tikhomirov:
Fix for MPS-38286 revealed few missing dependencies in the build project, the patch adds these into com.mbeddr.platform build project


sergej-koscejev (approved):
I don't like one added dependency on jetbrains.mps.testing but I will approve this PR in the interest of having a new MPS release that will give us the diagnostics we need to adjust dependencies properly.

PR #2935, opened 2025-05-07
Merge maintenance/mps20251-2934 into master

mps-ci-buildbot:
Merge changes from `maintenance/mps20251-2934` to `master`.


sergej-koscejev:
@fisakov please don't use `maintenance/` prefix for your own branches since they will get involved in our cascading merges and we don't want that.

fisakov:
> @fisakov please don't use `maintenance/` prefix for your own branches since they will get involved in our cascading merges and we don't want that.

@sergej-koscejev if your workflow is so sensitive, you should setup [branch protection rules](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule)

sergej-koscejev:
@fisakov I prefer to talk to people first, and set up technical measures only if that doesn't work.

PR #2942, opened 2025-05-13
Fix test.ts.ext.components.embedded to work with 2025.1

sergej-koscejev:
Add a build configuration to ensure that test classes are not dropped.


wsafonov (approved):

fisakov:
When is this going to be merged in master? Our tests in 2025.1 are running smoothly now, but master BC, which is based on mbeddr's master, is failing.

Issue #2958, opened 2025-05-19
mbeddr.doc: instable SIDs in Table of Content after several generations


arimer:
Found in maintenance/mps20232

When exporting documents with a `TableOfContents` holding instances of `DocumentInclude` (withtouth the `referenceOnly` flag set), the generated html links to the individual parts of the TOC don't keep stable SIDs. On each generation new IDs are created resulting in instable links over multiple generations.

**Header and div IDs change**
- when re-generating [MainDoc](http://127.0.0.1:63320/node?ref=r%3Ac20abc99-c3e1-4a29-9270-de587f318ef3%28com.mbeddr.doc.test.documents.doc_refs%29%2F2671071626003091836) the individual IDs are no longer stable. 
<img width="2435" alt="Image" src="https://github.com/user-attachments/assets/1f9b051b-ba0e-46b7-bf94-7d02302e7340" />

arimer:
I am not sure how this bug could stay hidden for so long, but in MHO we should use node IDs from nodes inside the `original input model ` when calculating SIDs instead of using the current node in the current transient model like it is implemented [now](http://127.0.0.1:63320/node?ref=r%3Aeedc51db-c07c-4a6b-a018-0a59455fef83%28com.mbeddr.doc.gen_xhtml.generator.template.main%40generator%29%2F5185579450379235342)

PR #2959, opened 2025-05-19
mbeddr.doc: make IDs robust over several generations

arimer:
This PR addresses https://github.com/mbeddr/mbeddr.core/issues/2958 and uses original node IDs for generating SIDs (stable IDs) where possible. 

In addition some rudimental tests were added to check for generation of stable IDs.

**Explanations**
Previously, only node.StableID() was utilized. By default, the behavior method is applied to the node within the generator model (transient). The node ID is updated every time because the generator always creates a copy of all nodes and transfers them to the transient model.

To obtain **real** stable IDs, we need to use the IDs from the original model. However, in some cases, a node may not have been present in the original model because it was previously transformed or created in another generator. Initially, I did not include an “instance of” check when accessing the original node, but it has been identified that this can lead to unexpected results.

The “instance of” check serves as a safety net. If an original instance exists, it is used. Otherwise, the same logic as usual is applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 # May 2025
 
+## com.mbeddr.doc
+- Use original nodeID for stable ID (SID) generation instead of transient nodeID
+
 ## com.mbeddr.mpsutil.actionsfilter
 
 - An exception (!app.isDispatchThread()) was fixed that was related to the initialization of the actionsfilter language.
--- a/code/languages/com.mbeddr.doc/languages/com.mbeddr.doc.gen_latex/generator/template/com/mbeddr/doc/gen_latex/generator/template/main@generator.mps
+++ b/code/languages/com.mbeddr.doc/languages/com.mbeddr.doc.gen_latex/generator/template/com/mbeddr/doc/gen_latex/generator/template/main@generator.mps
@@ -25,6 +25,7 @@
     <import index="tpck" ref="r:00000000-0000-4000-0000-011c89590288(jetbrains.mps.lang.core.structure)" />
     <import index="wyt6" ref="6354ebe7-c22a-4a0f-ac54-50b52ab9b065/java:java.lang(JDK/)" />
     <import index="48kf" ref="r:5f41c82d-84d1-4fb1-a1cf-6697d2365854(com.mbeddr.mpsutil.filepicker.behavior)" />
+    <import index="znf5" ref="r:07597124-beb3-41b7-beb1-a882af3ded40(com.mbeddr.doc.plugin)" />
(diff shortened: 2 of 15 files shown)


nkoester:
There are quite a lot of repetitive calls such as

image

Maybe a static function in the [GeneratorHelper](http://127.0.0.1:63320/node?ref=r%3Aa38f0e6e-eb07-4712-927b-30c9e021556c%28com.mbeddr.doc.gen_xhtml.plugin%29%2F988357225313169968) could reduce this and the potential for errors.

arimer:
@nkoester done. I created a helper in mbeddr.doc.gen and made it re-usable among all other generator langauges that use this functionality.

sergej-koscejev:
> Tbh. I have no idea why I needed to do the change in [1825381](https://github.com/mbeddr/mbeddr.core/commit/1825381d20708be17371286c76e9128068a8f73a). It might be related to the bug that @sergej-koscejev found: https://youtrack.jetbrains.com/issue/MPS-38286/Build-language-does-not-detect-plugin-dependency-from-used-languages

There was a bug in the compare language which should be fixed by now and adding .behavior dependency should not be necessary.

sergej-koscejev:
@arimer why do we check the concept of the original node? Can't we use its ID for the stable ID? Assuming we only generate one thing out of that node, but if we generate several, the current implementation is also not going to work, is it?

arimer:
>The current implementation still relies on the transient IDocReferenceable in some cases and that doesn't really have a stable ID, does it?

This is correct. 

> What if DocGeneratorHelper.getOriginalStableId() would always get the original node and ask it for its stable ID if it's a IDocReferenceable, otherwise use the StableIdHelper to produce the ID based on the original node ID? IDocReferenceable would in effect be used for "customizing" the ID if the standard approach is not suitable.

I like the idea. In this case we do not rely on the "unstable IDs" for nodes from transient models. I'll update the code accordingly. ✅ 

>Also, generator context has a 'unique name' operation, could that help here?

I dont think so. The goal is to generate the same SIDs for the same "nodes" in the document. As far as I know, the `unique-name` will only create unique names during a generation run. The next generation will again produce (different) but unique IDs.

> Perhaps we should pass the transient node and the generation context to IDocReferencable#stableId()?

Not sure about that. How would it help?

sergej-koscejev (commented):
Please revert commit c3b8c1c90c600d598cab01ce9b9477e27ed35331 and squash the other commits (the PR has 20 commits now and it is a bit excessive).

Typos:
1. `sectionSHouldHaveStableId` → `sectionShouldHaveStableId`
2. The thrid section → The third section

Otherwise looks good, thanks!

sergej-koscejev (dismissed):

sergej-koscejev (approved):

PR #3013, opened 2025-06-01
base: prevent exception due to null idea project event parameter

wsafonov:
Fixed exceptions related to the search viewer (https://youtrack.jetbrains.com/issue/MPS-38663).

Original reported exception were not reproducible, but other exceptions due to null value of the idea project event parameter were thrown.


slisson (approved):

sergej-koscejev:
@wsafonov please add a changelog entry

PR #3018, opened 2025-06-02
sbom: include only runtime configs to the sbom

wsafonov:
Generated platform SBOM contains now only the runtime dependencies: bundled libraries, language libraries, mps and jbr.


wsafonov:
> What are runtime configurations and which configurations were included previously?

it just contained all configurations previously, i.e. also test or build-time dependencies, which should not be part of the SBOM. Runtime covers basically everything this library requires to run, not just the directly bundled libs.

sergej-koscejev (dismissed):
LGTM, please add the explanation about runtime configurations to the changelog.

PR #3022, opened 2025-06-04
ProjectFactory now gives MPSProject right away

artem-tikhomirov:
`MPSProject` soon to cease being IDEA `ProjectComponent`, and we strive to hide `ideaProject.getComponent(MPSProject.class)`, especially when there's no need, like in `ProjectFactory` case, where we know we create MPSProject.


sergej-koscejev:
@artem-tikhomirov FYI the build failed for this PR.

PR #3057, opened 2025-07-18
core.base: Update the RefreshF5 action in the background

alexanderpann:
- fixes https://github.com/mbeddr/mbeddr.core/issues/3039


sergej-koscejev (approved):

alexanderpann:
The change is always reverted by the re-runnable migrations. I added a dummy Frame parameter to the action so that the EDT access is required and the flag is not changed anymore.

Issue #2814, opened 2025-03-18
actionsfilter should activate default profile on first start of the RCP


sergej-koscejev:
It was observed that when an RCP that uses actionsfilter is started for the first time, the 'active by default' profile is not activated. I believe the fix would be to implement `noStateLoaded()` in `ActionsApplicationComponent` appropriately.

PR #3068, opened 2025-07-24
mpsutil.actionsfilter: Rework the language

alexanderpann:
- make code simpler
- Get rid of the IntelliJ plugin and use an lightweight IntelliJ service instead. When you want to modify the runtime solution, you only have to restart MPS because the service can't be reloaded.
- only save necessary information in actionFilter.xml (e.g. don't save default profiles)
- UI improvements
- add a sandbox
- fixes https://github.com/mbeddr/mbeddr.core/issues/2814

Bildschirmfoto 2025-07-29 um 12 35 23


sergej-koscejev:
I haven't looked at the code yet but I welcome the initiative.

IntelliJ Platform has a class called [CustomActionsSchema](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/ide/ui/customization/CustomActionsSchema.kt) which also stores action customizations. Maybe we could also eventually rework the runtime to make use of that functionality.

alexanderpann:
I didn't know about this class, but I am not sure how it would help. The extension from this class allows adding an action group to the list of customizable actions in Settings | Menus and Toolbars:

Bildschirmfoto 2025-07-25 um 08 31 15

What should be probably reworked is the hard-coded list top level main menu groups like "File" and "Edit" that we have in the code. There must be a better way to populate this list.

AtanasM98 (approved):

sergej-koscejev:
@alexanderpann setActions ,ethos of this class sets action customizations, i.e. added/deleted actions for various places.

sergej-koscejev:
@alexanderpann @AtanasM98 a few findings (the PR was merged before I could review it):
1. I like the GUI changes, thanks for taking time to improve the UI.
2. Is it necessary to call the model `com.mbeddr.mpsutil.actionsfilter.ideaplugin.plugin`? Wouldn't just `.ideaplugin` suffice? Or doesn't IDEA see it then?
3. Since you are moving the ActionsApplicationComponent class anyway, you could rename it to ActionsService (it's no longer an application component but I didn't rename it because of compatibility).
4. Speaking of compatibility, moving or renaming this class breaks binary compatibility with previously compiled action filter plugins, due to `ActionsApplicationComponent` being used in the generator. If this cannot be avoided, it should at least be clearly stated in the changelog.
5. "Copy Profile" action should be called "Duplicate Profile" ("Copy" is for copying to the cliipboard).

PR #3069, opened 2025-07-29
mpsutil.actionsfilter: ensure that default profiles are always loaded (follow-up)

alexanderpann:
This is a follow-up of https://github.com/mbeddr/mbeddr.core/pull/3068#issuecomment-3130659610:

2. Fixed
3. Fixed, I didn't see the "Application" part.
4. I moved it on purpose, so that we don't get ClassCastExceptions when MPS tries to reload the service when you build the runtime solution, and
 you have to restart MPS. It doesn't work at the moment but maybe MPS is smart enough to rebuild it in a future version.
5. Fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,6 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 ## com.mbeddr.mpsutil.actionsfilter
 
 - The actionsfilter plugin now correctly initializes when opening an RCP for the first time and also loads the default actions profile. Various other improvements were made to the language.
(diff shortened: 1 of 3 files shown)


sergej-koscejev:
I can have a look at the binary compatibility but probably only in two days or so.

alexanderpann:
Ok. I am not even sure what getInstance should return because we don't have an application component anymore.

sergej-koscejev:
@alexanderpann see commit 0e9519f. Let me know what you think.

alexanderpann:
@sergej-koscejev Looks good to me. You can approve the PR.

sergej-koscejev (approved):

PR #3078, opened 2025-08-07
mpsutil.spreferences: rework SPreferencePage class loading

alexanderpann:
The associated model with the preference page now also is not null anymore and the exception about the displayName should be finally fixed.


alexanderpann:
The original issue was similar to https://youtrack.jetbrains.com/issue/MPS-36573/No-display-name-specified-in-plugin-descriptor-XML-file-for-configurable.  I thought that I already added a changelog entry the last time I tried to fix it.
I'll make those changes in the next PR. The dummy application plugin is necessary, so that I can use an existing plugin ID that I can use for the plugin descriptor.

sergej-koscejev:
> The dummy application plugin is necessary, so that I can use an existing plugin ID that I can use for the plugin descriptor.

@alexanderpann please put a comment somewhere in the plugin to explain that. E.g. create an empty init block or something.

PR #3087, opened 2025-08-14
Improve copying of custom cells and editors with custom swing components to plain text

alexanderpann:
I am talking about the feature that you can press ctrl/cmd + c and ctrl/cmd +v to paste nodes from the editor to the clipboard or a normal text file.


sergej-koscejev (changes requested):
I don't think we should override toString() of components in the component cell, this looks somehow adhoc and ugly and I don't like doing it for every Swing component cell in the project just in case somebody would want to copy-paste things.

I think we should talk to JetBrains and ask them to add a query function to component cells that would return their text presentation. Or perhaps create our own Swing component cell with such a query function first.

Issue #3100, opened 2025-08-27
[mbeddr.doc] Render content of ModelContentAsTextParagraph in generated document


hubermanuel:
If you use "embed model content as text" the generated document don't show the content directly. You can achieve this by using the intention "Update Model Content Text". This intention copies the content of the linked node into the text property. The text property is then used when rendering the document.
However, this behavior is highly error prone due to the fact that you need to use this intention every time you change something in the linked node.

PR #3101, opened 2025-08-27
mbeddr.doc: Fix embed model content as text #3100

hubermanuel:
This PR tackels the ticket #3100 
I implemented the functionality to render the text of the linked node of a ModelContentAsTextParagraph directly in the generator. After this change a user no longer needs to update the text property with the intention "Update Model Content Text".

I'm not sure if the intention UpdateCodeText is still needed/wanted. If some project leads confirms this, I will add a commit to delete it directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 # August 2025
 
+## com.mbeddr.doc
+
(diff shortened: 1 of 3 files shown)


sergej-koscejev (changes requested):
Requesting changes, see my previous comment.

alexanderpann:
@sergej-koscejev @hubermanuel There was no update in the last 3 months. Should we close the PR?

PR #3118, opened 2025-09-08
build: com.mbeddr.mpsutil.plantuml: replaced old httpsupport with correct one

arimer:
We had still an outdated plugin dependency to  `com.mbeddr.mpsutil.httpsupport`. I replaced it with `jetbrains.mps.ide.httpsupport`.


slisson (commented):
Looks like it was intentionally changed from `jetbrains.mps.ide.httpsupport` to `com.mbeddr.mpsutil.httpsupport` before. See https://github.com/mbeddr/mbeddr.core/commit/e64bcea36479aa64e297e59a15a2e34248228865

PR #3128, opened 2025-09-10
build(platform): add provided dependencies of mpsLibraries to the POM

sergej-koscejev:
These are the libraries that are bunded with MPS-extensions, and therefore with mbeddr platform too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ Semantic Versioning and the changes are simply documented in reverse chronologic
 
 # September 2025
 
-- The POM file of the mbeddr platform now includes bundled dependencies with 'provided' scope.
+- The POM file of the mbeddr platform now includes bundled dependencies with 'provided' scope, including the
+  bundled dependencies of MPS-extensions. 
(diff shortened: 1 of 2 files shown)


slisson:
What's the benefit of having these dependencies with `provided` scope in the pom?

slisson:
I'm not an expert on that, but I can see that there already is an sbom.json packaged into the zip file for exactly this purpose.

If I understand it correctly, this sbom.json is generated from the maven dependencies, but adding dependencies to the maven pom just for the purpose of them being listed in the sbom looks like a misuse of the maven dependencies.

Does it make sense to modify the sbom directly?

Since there are already other dependencies with provided scope in the pom, this PR just makes the list complete and whether that's the right place for them or not is a different discussion. We can merge this change for now and think about alternatives later.

sergej-koscejev:
The customer's tool is actually looking for the dependencies in Maven POMs, for some reason it was easier for them than to look at SBOMs (although they did start with using SBOMs initially). I will check with them to better understand why.

PR #3166, opened 2025-09-27
build: compute version info once per build

sergej-koscejev:
* Move version computation to a settings plugin.
* Share version info in beforeProject lifecycle callback (Gradle best practice compliant with project isolation).
* Remove unused mpsExtensionsVersion property.
* Store MPS version in gradle.properties.


sergej-koscejev:
2023.2 is too old, I'm going to rebase it on 2024.1 or some later version.

PR #3178, opened 2025-10-06
actionsfilter: fix exceptions at startup

sergej-koscejev:
Only remove actions from DefaultActionGroups, treat other ActionGroups as opaque. Use `getChildActionsOrStubs()` to access child actions and avoid stack traces about `getChildren(null)`.


arimer (approved):
Changes look good. Thanks 👍🏼

sergej-koscejev:
> Will this also fix the exceptions when opening the actions filter preference page via the settings menu?

Yes, they are all caused by calling `ActionGroup#getChildren(AnActionEvent)` and I believe I have eliminated all such calls.

PR #3208, opened 2025-10-20
Restore the changelog

alexanderpann:
Some entries got lost.


sergej-koscejev (changes requested):
This adds another entry for mbeddr.doc to the changelog, this should be changed.

sergej-koscejev (approved):

PR #3212, opened 2025-10-20
renovate: use custom versioning scheme for MPS-extensions

sergej-koscejev:
The default versioning scheme treats commit hashes with `a` or `b` in them as pre-releases. The custom scheme only treats `-something` suffix as a pre-release.


sergej-koscejev:
Bypassing merge checks because changes are only in Renovate files.

PR #3432, opened 2026-05-05
fix QueryStringEncoder dependency

mvlassiev:
QueryStringEncoder is no longer coming from io.netty solution but is instead resolved via MPS.IDEA.
 (It appears that the io.netty solution is no longer used.)


sergej-koscejev:
@mvlassiev please revert the resolve info changes, they are breaking the build.

PR #3498, opened 2026-08-20
Mps/2026.2

fisakov:
These are changes necessary to have mbeddr built with MPS 2026.2.


sergej-koscejev:
Need to fix the FIXME before this can be merged.

PR #3519, opened 2026-09-05
fix: defer MbeddrSearchViewer tool-window registration until safe

danielratiu:
## Problem

Opening a project in mbeddr.formal (and any project which uses mpsutils) throws:

```
java.lang.IllegalStateException: You must not register toolwindow programmatically so early. Rework code or use ToolWindowManager.invokeLater
    at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.getDefaultToolWindowPaneIfInitialized(ToolWindowManagerImpl.kt:490)
    at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.registerToolWindow(ToolWindowManagerImpl.kt:968)
    ...
    at jetbrains.mps.ide.tools.BaseTool.getToolWindow(BaseTool.java:199)
    at c.m.c.base.pluginSolution//com.mbeddr.core.base.pluginSolution.plugin.MbeddrSearchViewer_Tool.lambda$setProject$0(MbeddrSearchViewer_Tool.java:52)
    ...
```

## Root cause

`MbeddrSearchViewer_Tool.init(project)` unconditionally calls `this.setProject()` right after constructing the search-viewer component — this runs automatically the moment the tool is constructed during project startup, which can be the very first access to this tool's `ToolWindow`.

`setProject()` deferred its `getToolWindow().setTitle(...)` call via bare `ApplicationManager.getApplication().invokeLater(...)`. That only posts to the next EDT event — it doesn't wait for the tool-window infrastructure to be ready. On current IntelliJ platform versions that deferred callback can still run before the frame's default tool-window pane is initialized, so the lazy `getToolWindow()` → `register()` → `ToolWindowManager.registerToolWindow()` call throws.

`setGlobal()`, `setModel()`, and `setModule()` have the identical `ApplicationManager.getApplication().invokeLater(...)` pattern, though only `setProject()` is actually reachable from an unconditional startup call — the other three are only invoked from user-triggered "search" actions, which already call `getToolWindow()`/`openTool()` synchronously first (so registration has already happened by the time they run). Fixing all four removes any reliance on that call-ordering coincidence, at no behavioral cost.

## Fix

Replace `ApplicationManager.getApplication().invokeLater(...)` with `ToolWindowManager.getInstance(project).invokeLater(...)` in all four methods — the mechanism the platform's own exception message points to. It defers the callback until tool-window registration is actually safe, instead of just "next EDT tick".

## Scope: other `ApplicationManager...invokeLater` + `getToolWindow()` usages

Searched all `BaseTool.getToolWindow()` call sites across mbeddr.core. Besides this fix:

- `GenericTreeTool`, `com.mbeddr.mpsutil.graphstream`, `com.mbeddr.mpsutil.plantuml.pluginSolution` — no `isAvailableOnStartup`, `getToolWindow()` only reachable from explicit "open tool" user actions. Not at risk.
- `com.mbeddr.mpsutil.contextactions.runtime` (`ContextActionsTool`, `isAvailableOnStartup: true`) — already uses the correct `ToolWindowManager.getInstance(project).invokeLater(...)` pattern; good independent confirmation this is the right idiom. Its other `getToolWindow()` call site is reached only via a `ToolWindowManagerListener.toolWindowShown(...)` callback, which by construction can't fire before the tool-window system is already up.
- `com.mbeddr.doc.aspect.ui` (`TabbedDocViewer`, `isAvailableOnStartup: true`) — no `init()` override; its `getToolWindow()`/`register()` call is reached only from editor/selection-change listeners, not from an unconditional startup call. Verified empirically too: opening the safety project triggers neither this nor the contextactions path.

None of these have `MbeddrSearchViewer`'s specific shape — an **unconditional call from `init(project)`** into `getToolWindow()`. That's the actual distinguishing risk factor, not `isAvailableOnStartup` or the mere presence of `ApplicationManager...invokeLater`.

## Testing

Verified in MPS: all four methods type-check cleanly (`check_root_node_problems` clean) and a model make/generate completed with no code-level errors. Not yet re-tested against the live crash in mbeddr.formal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/code/platform/com.mbeddr.mpsutil/solutions/com.mbeddr.core.base.pluginSolution/com.mbeddr.core.base.pluginSolution.msd
+++ b/code/platform/com.mbeddr.mpsutil/solutions/com.mbeddr.core.base.pluginSolution/com.mbeddr.core.base.pluginSolution.msd
@@ -46,6 +46,7 @@
     <language slang="l:a247e09e-2435-45ba-b8d2-07e93feba96a:jetbrains.mps.baseLanguage.tuples" version="1" />
     <language slang="l:63650c59-16c8-498a-99c8-005c7ee9515d:jetbrains.mps.lang.access" version="0" />
     <language slang="l:aee9cad2-acd4-4608-aef2-0004f6a1cdbd:jetbrains.mps.lang.actions" version="4" />
+    <language slang="l:af65afd8-f0dd-4942-87d9-63a55f2a9db1:jetbrains.mps.lang.behavior" version="2" />
     <language slang="l:ceab5195-25ea-4f22-9b92-103b95ca8c0c:jetbrains.mps.lang.core" version="2" />
     <language slang="l:18bc6592-03a6-4e29-a83a-7ff23bde13ba:jetbrains.mps.lang.editor" version="15" />
     <language slang="l:c0080a47-7e37-4558-bee9-9ae18e690549:jetbrains.mps.lang.extension" version="2" />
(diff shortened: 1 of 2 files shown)


sergej-koscejev:
Need to verify whether this is a problem in 2024.1 and 2025.1 as well. Would be nice if your agent could do this @danielratiu.

